PR #7, opened 2015-02-18
Changes to build against spark-1.3

antonycourtney:
Minimal changes to get this building with spark-1.3.0-SNAPSHOT from apache/spark github repo.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,8 +10,11 @@ scalaVersion := "2.10.4"
 
 libraryDependencies += "org.apache.hadoop" % "hadoop-client" % "1.0.4"
 
-libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.2.0"
+libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0-SNAPSHOT"
 
 libraryDependencies += "com.google.guava" % "guava" % "14.0.1" % Test
 
 libraryDependencies += "org.scalatest" %% "scalatest" % "2.1.5" % Test
+
+resolvers += Resolver.mavenLocal
+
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/examples/redshift/input/RedshiftInputFormat.scala
```diff
@@ -104,7 +106,7 @@ object RedshiftInputFormat {
       import sqlContext._
       val structType = SchemaParser.parseSchema(schema)
       val casts = structType.fields.map { field =>
-        field.name.attr.cast(field.dataType).as(Symbol(field.name))
+        col(field.name).cast(field.dataType).as(Symbol(field.name))
```
rxin: you can actually remove Symbol here

src/main/scala/com/databricks/examples/redshift/input/SchemaParser.scala
```diff
@@ -43,7 +44,7 @@ private[redshift] object SchemaParser extends JavaTokenParsers {
     case colName ~ colType => StructField(colName, colType, nullable = true)
   }
   private val structType: Parser[StructType] = structField.* ^^ {
-    case fields => new StructType(fields)
+    case fields => new StructType(fields.toArray)
```
rxin: you can do

```scala
StructType(fields)
```

(i.e. without the new)

build.sbt
```diff
@@ -10,8 +10,11 @@ scalaVersion := "2.10.4"
 
 libraryDependencies += "org.apache.hadoop" % "hadoop-client" % "1.0.4"
 
-libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.2.0"
+libraryDependencies += "org.apache.spark" %% "spark-sql" % "1.3.0-SNAPSHOT"
```
JoshRosen: Maybe this could depend on the 1.3.0 RC1 published to the Apache staging repo instead, since that won't require publishLocal?

JoshRosen: Actually, that might not be a great idea since the snapshot artifacts have the version "1.3.0" even though they might not reflect the final 1.3.0 that gets released, so this could cause RCs to get cached in user Maven / Ivy caches and stay there even after 1.3.0 final is released.

PR #9, opened 2015-03-16
Add support for 'text' fields

nburoojy:
Currently, spark-redshift assumes all text types are followed by the length in parens. The length is not required for redshift types, so I made it optional.


mengxr:
Close & re-open to trigger Travis.

mengxr:
LGTM.

PR #12, opened 2015-03-23
add .travis.yml

mengxr:


JoshRosen:
What about caching the Ivy directories, like we did for `spark-perf`?

```
cache:
  directories:
    - $HOME/.ivy2
```

PR #19, opened 2015-05-20
make hadoop version configurable

mengxr:
Package built with Hadoop 1.x is not compatible with Hadoop 2.x. So we need to make Hadoop version configurable and release multiple artifacts.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 net.virtualvoid.sbt.graph.Plugin.graphSettings
 
+val hadoopVersion = settingKey[String]("Hadoop version")
+
 organization := "com.databricks"
 
 name := "spark-redshift"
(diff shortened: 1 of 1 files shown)


JoshRosen:
Do you think that we could work around this incompatibility using reflection instead? If a single TaskAttemptContext method is the only source of incompatibility then we might be able to do something similar to https://github.com/apache/spark/pull/6599 to work around this.

JoshRosen:
Also, minor complaint: it would have been really helpful to add a comment to `build.sbt` explaining why the version needs to be configurable, since I think that would have prevented this change from being accidentally undone in a later PR.

mengxr:
I don't remember what was the incompatibility issue here. If it is only `TaskAttemptContext` and we could use reflection, that's good. But it is still useful to test against multiple hadoop verisons in Travis.

Issue #8, opened 2015-03-09
Feature : Add write feature in redshift through spark-core and spark-sql


mkanchwala:
There should be a way to write the RDDs to Redshfit in Spark with some example

mengxr:
We are going to implement the data source API with read/write support. I'll keep you updated.

Aerlinger:
Any updates on this? I'm looking to implement this myself and would rather join forces than duplicate work. Happy to help out in any way I can.

bumrush:
I would like to know about this too. Is there a way to write RDDs to Redshift?

mengxr:
Sorry, we haven't had time to implement this feature. The current plan is to make our design doc public and see if there are people who are interested in working on it. I will add the link soon.

jaley:
Hi there!

We (Analytics team at SwiftKey) have planned some capacity in our next sprint to get our Spark jobs writing data out to Redshift. Were were planning to do something crude with an S3 CSV export, but the better way for it to work would probably be by adding the write support to this library. If you're able to provide some support, perhaps on the mailing list, this thread, IRC, or whatever you prefer to use, we'd be happy to have a go at implementing this feature and contributing the code back.

Are you able to share that design document you mentioned? We're likely to need a bit of guidance, and that would be a good start.

Also, we have a bunch of things that we'd like this feature to involve to make our ETL pipeline work the way we need. Hopefully these are generally desirable features for the write functionality, but here's our wish list, which we'd hope to include in the implementation, for wider review:

* DataFrames can be saved to a Redshift table via `save()` function (think that's pretty core!)
* Automatically infers a 'default' Redshift schema from the DataFrame column types, but it should be possible to provide more specific overrides for these with options somehow. For example, a DataFrame `StringType` column may sometimes needed to be written as `TIMESTAMP WITHOUT TIMEZONE NOT NULL` in Redshift.
* Ability to specify distribution key, sort keys, etc.
* Data load should be done via S3, with potentially configurable, but at least sensible default data encodings, compression, etc. We could potentially make use of the new Avro support in Redshift, if there are performance advantages (I'm not sure). We don't actually require support for loading data via SQL `INSERT` commands, and I imagine most people don't, as it's painfully slow and Amazon don't recommend it.
[...]

PR #22, opened 2015-07-15
Data Source API and JDBC/S3 read/write

jaley:
This is unfinished, I'll be updating the PR today and tomorrow with tests and documentation, but it's basically functionally complete and manual testing with Redshift directly suggests it pretty much works, so I thought it'd be a good time to start gathering feedback.

Included in this PR:

* Data Source API implementation, including filtering/pruned traits, insertable, create, etc. `DefaultSource` for integration with Spark SQL too.
* Read and write support, using S3 for data transfer and JDBC to initiate loads, unloads, etc.
* Makes use of the new Avro support in Redshift (requires Redshift version 1.0.963+) for writes. Redshift currently doesn't support Avro for its `UNLOAD` command.
* Some useful Redshift functionality for setting distribution and sort keys, plus some good ETL practices such as staging tables etc.

Issue #8 has a thread with a few limitations and challenges I'm still looking for better solutions to. Please feel free to take this for a spin and make suggestions here!


mengxr:
ping @kai-zeng

jaley:
@mengxr - in response to your questions on the other thread:

Avro "seems" faster, but I'll do a more controlled test when I get a chance. I think the main benefit is that it avoids having to mess around with delimiters and escape characters though. I foresee bug reports where someone has some kind of unexpected character/quote/etc in their data that breaks whatever text-delimited serializer we picked instead of Avro. If Amazon add support for unloading with Avro too, it'd be great to switch to that later and drop the text-based formats altogether.

jaley:
Hey @mengxr - I'm having some Travis drama here. This builds and tests fine for me, using same Scala, Spark and Hadoop versions, but Travis always fails with this shutdown hook priority exception (stack trace is in the Travis logs). Any suggestions? I've no idea what's up with it.

Test coverage is up at 82% now, I think a few more cases should get us there. I'll add docs on Monday.

mengxr:
I saw the following. It looks like a dependency issue. Could you check the dependency from `spark-avro_2.10:1.0.0`?

~~~
java.lang.NoClassDefFoundError: Could not initialize class org.apache.spark.util.Utils$
~~~

jaley:
@mengxr Found the problem -- seems there were conflicting hadoop dependencies on the classpath, as usual. I cleaned up the dependencies to fix it.

I need to improve coverage a little more, but tests are now passing except for two of the original input format tests. These only fail on Travis, they pass on my laptop locally, and I'm not sure what they mean. The two assertions causing the failure are the ones commented out in this commit:

https://github.com/databricks/spark-redshift/commit/282075f029e1e1eff4b07b8ed40c814197161db2

I don't understand why we would want there to be more partitions than records? I especially don't understand why that would be true on my laptop but not on Travis either! I think the main difference now is that I'm on Java 8 and Travis is on Java 6. I just left them commented for now while I fix other things, but it would be good to understand what's going on there.

jaley:
@mengxr I've added some documentation to explain how to use the new functionality, and how the parameters work. See updated `README.md` file.

This is pretty much ready to go -- just those remaining test issues to fix, which I need some help with, as I'm not familiar with the input format code. I think it's ready for a review anyway.

README.md
```diff
@@ -1,60 +1,128 @@
-RedshiftInputFormat
-===
+# RedshiftInputFormat
```
marmbrus: spark-redshift?

README.md
```diff
@@ -1,60 +1,128 @@
...
-Otherwise,
+// These three settings are required, see below for full information
+val params = Map("jdbcurl" -> "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass",
+                 "redshifttable" -> "my_table",
+                 "tempdir" -> "s3://path/for/temp/data")
+
+// Get some data from a Redshift table
+val df: DataFrame = sqlContext.read
+    .format("com.databricks.spark.redshift")
+    .options(params)
```
marmbrus: I'd consider using the following syntax (in general I try to use syntax that works across languages wherever possible in example code)

```
.option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass")
.option("table", "my_table")
```

README.md
```diff
@@ -78,12 +146,153 @@ val records: DataFrame = sqlContext.redshiftFile(path, Seq("name", "age"))
...
+<table>
+ <tr>
+    <th>Parameter</th>
+    <th>Required</th>
+    <th>Default</th>
+    <th>Notes</th>
+ </tr>
+
+ <tr>
+    <td><tt>redshifttable</tt></td>
```
marmbrus: `dbtable` to match the JDBC data source?

README.md
```diff
@@ -78,12 +146,153 @@ val records: DataFrame = sqlContext.redshiftFile(path, Seq("name", "age"))
...
+    <td>Yes</td>
+    <td>No default</td>
+    <td>The table to create or read from in Redshift</td>
```
marmbrus: Does this also support a full query in parentheses?

README.md
```diff
@@ -78,12 +146,153 @@ val records: DataFrame = sqlContext.redshiftFile(path, Seq("name", "age"))
...
+ </tr
+ <tr>
+    <td><tt>jdbcurl</tt></td>
```
marmbrus: Maybe we should make this `url` so that it matches up with the JDBC data source?

README.md
```diff
@@ -78,12 +146,153 @@ val records: DataFrame = sqlContext.redshiftFile(path, Seq("name", "age"))
...
+    <td><tt>jdbcdriver</tt></td>
+    <td>No</td>
+    <td><tt>org.postgresql.Driver</tt></td>
+    <td>The class name of the JDBC driver to load before JDBC operations. Must be on classpath.</td>
+ </tr>
+ <tr>
+    <td><tt>overwrite</tt></td>
+    <td>No</td>
+    <td><tt>false</tt></td>
+    <td>If true, drop any existing data before writing new content. See also <tt>usestagingtable</tt></td>
```
marmbrus: Why do we need this property in addition to the save mode?

src/main/scala/com/databricks/spark/redshift/Conversions.scala
```diff
@@ -0,0 +1,130 @@
...
+  val TRUE: Parser[Boolean] = "t" ^^ Function.const(true)
+  val FALSE: Parser[Boolean] = "f" ^^ Function.const(false)
+
+  def parseRedshiftBoolean(s: String): Boolean = parse(TRUE | FALSE, s).get
+}
+
+/**
+ * Data type conversions for Redshift unloaded data
+ */
+object Conversions {
```
marmbrus: `private[redshift]`

src/main/scala/com/databricks/spark/redshift/Conversions.scala
```diff
@@ -0,0 +1,130 @@
...
+   * Convert schema representation of Dates to Timestamps, as spark-avro only works with Timestamps
+   */
+  def datesToTimestamps(sqlContext: SQLContext, df: DataFrame): DataFrame = {
+    val schema = StructType(
+      df.schema map {
+        case StructField(name, DateType, nullable, meta) => StructField(name, TimestampType, nullable, meta)
+        case other => other
+      })
+
+    sqlContext.createDataFrame(df.rdd, schema)
```
marmbrus: We should probably use a cast here instead of just changing the schema (in case internal details about the difference between date and timestamp ever change).

src/main/scala/com/databricks/spark/redshift/DefaultSource.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  override def createRelation(sqlContext: SQLContext, parameters: Map[String, String], schema: StructType): BaseRelation = {
+    val params = Parameters.mergeParameters(parameters)
+    RedshiftRelation(jdbcWrapper, params, Some(schema))(sqlContext)
+  }
+
+  /**
+   * Creates a Relation instance by first writing the contents of the given DataFrame to Redshift
+   */
+  override def createRelation(sqlContext: SQLContext,
```
marmbrus: Wrap parameters with 4-space indent

src/main/scala/com/databricks/spark/redshift/Parameters.scala
```diff
@@ -0,0 +1,183 @@
...
+
+      val (accessKeyId, secretAccessKey) =
+        if(parameters.contains("aws_access_key_id")) {
+          (parameters("aws_access_key_id"), parameters("aws_secret_access_key"))
+        } else {
+          try {
+            val awsCredentials = (new DefaultAWSCredentialsProviderChain).getCredentials
+            (awsCredentials.getAWSAccessKeyId, awsCredentials.getAWSSecretKey)
+          } catch {
+            case e: Exception => throw new Exception("No credentials provided and unable to detect automatically.", e)
```
marmbrus: What about when the credentials are specified in the URL that is given or in the hadoop configuration?

src/main/scala/com/databricks/spark/redshift/RedshiftInputFormat.scala
```diff
@@ -20,13 +20,13 @@ import java.io.{BufferedInputStream, IOException}
...
-import scala.collection.mutable.ArrayBuffer
-
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.io.compress.CompressionCodecFactory
-import org.apache.hadoop.mapreduce.{InputSplit, RecordReader, TaskAttemptContext}
 import org.apache.hadoop.mapreduce.lib.input.{FileInputFormat, FileSplit}
+import org.apache.hadoop.mapreduce.{InputSplit, RecordReader, TaskAttemptContext}
+
+import scala.collection.mutable.ArrayBuffer
```
marmbrus: This should actually go back (https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide#SparkCodeStyleGuide-Imports)

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -0,0 +1,131 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.databricks.spark.redshift
+
+import java.util.Properties
+
+import com.databricks.spark.redshift.Parameters.MergedParameters
```
marmbrus: This should go last in its own group.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.spark.sql.jdbc.JDBCWrapper
+import org.apache.spark.sql.sources._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+
+/**
+ * Data Source API implementation for Amazon Redshift database tables
+ */
+private [redshift]
+case class RedshiftRelation(jdbcWrapper: JDBCWrapper,
```
marmbrus: Wrap with 4space indent (https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide#SparkCodeStyleGuide-Indentation)

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -0,0 +1,131 @@
...
+    makeRdd(schema)
+  }
+
+  override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
+    val columns = columnList(requiredColumns)
+    unloadToTemp(columns)
+    makeRdd(pruneSchema(schema, requiredColumns))
+  }
+
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
```
marmbrus: You only need to implement this one (and you can remove the other Scan traits as they will never be called)

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -0,0 +1,131 @@
...
+    unloadToTemp(columns, whereClause)
+    makeRdd(pruneSchema(schema, requiredColumns))
+  }
+
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    val updatedParams = Parameters.mergeParameters(params.parameters updated ("overwrite", overwrite.toString))
+    new RedshiftWriter(jdbcWrapper).saveToRedshift(sqlContext, data, updatedParams)
+  }
+
+  def unloadToTemp(columnList: String = "*", whereClause: String = ""): Unit = {
```
marmbrus: `protected` here and all methods beneath.

marmbrus:
This is really awesome, thanks for working on it!

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def makeRdd(schema: StructType): RDD[Row] = {
+    val sc = sqlContext.sparkContext
+    val rdd = sc.newAPIHadoopFile(params.tempPath, classOf[RedshiftInputFormat],
+      classOf[java.lang.Long], classOf[Array[String]], sc.hadoopConfiguration)
+    rdd.values.map(Conversions.rowConverter(schema))
+  }
+
+  def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
+    val fieldMap = Map(schema.fields map { x => x.metadata.getString("name") -> x }: _*)
```
marmbrus: Can you explain what is going on with this function?  Why are you using the metadata name instead of just `.name` on the struct field?

README.md
jaley: Not tested, but I expect not, unfortunately, as I wasn't expecting it in the various places I reference the table name string from parameters. I mentioned this in the discussion on #8 but couldn't find a nice way to enable users to either work with a whole table directly or push some work out to Redshift that they know it'll handle well, then work with the resulting data set in a DataFrame.

Does the JDBC data source support that? I guess I missed it. Maybe we could raise an issue to add support for this in a separate change? It would be a nice a feature.

README.md
jaley: This is so that users can specify overwrite/append when using the `saveAsRedshiftTable` function in `package.scala`. I wasn't sure whether that public interface is really enough needed now that we're using data sources, but I put it there to match the existing input format API. For now, I just updated the docs to make it clear what this param is for, but we could potentially make this functionality data sources only if you prefer.

src/main/scala/com/databricks/spark/redshift/Conversions.scala
jaley: OK, not sure how to do this though. What would we be casting too? `java.sql.Timestamp`? If the objects in a `DateType` column of `df` here were not already `java.sql.Timestamp` instances, then how do I know how to convert them?

src/main/scala/com/databricks/spark/redshift/Parameters.scala
jaley: Do you mean something like `s3://user:pass@bucket/.../`? I'm pretty sure Redshift doesn't allow that in `COPY` or `UNLOAD`, as the syntax for those commands says `CREDENTIALS` is a mandatory parameter: http://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html#r_COPY-syntax-overview-credentials

I guess we could potentially parse the URI, extract the credentials and then reformat it without them or something? Getting complicated though!

I've added support for reading `fs.<scheme>.*` from the hadoopConfiguration, where scheme is determined by `tempdir`.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
jaley: Good question, I don't really get that either. I actually lifted this from the JDBC code because it was all private and I needed the exact same functionality. I've changed to using `x.name` directly and it's fine.

jaley:
@marmbrus Thanks for the review feedback! I think I've addressed all the comments, but for the few questions above.

src/main/scala/com/databricks/spark/redshift/Conversions.scala
marmbrus: Ah, sorry.  I meant a SQL cast.

```scala
val castedInput = df.schema map {
  case f if f.dataType == DateType => df(f.name).cast("timestamp").as(f.name)
  case f => df(f.name)
})
df.select(castedInput: _*)
```

README.md
marmbrus: It does if you put the query in `()` (so that its a subquery in the from clause).  We can add this as another feature.

marmbrus:
This is pretty awesome!  I propose that we merge this and then create issues for the following:
 - Clean up the coverage of the input format
 - Make sure we can handle full queries instead of just table names

After that we can make another release of this package.

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -0,0 +1,486 @@
...
+
+    val jdbcWrapper = prepareUnloadTest(params)
+    val testSqlContext = new SQLContext(sc)
+
+    // Assert that we've loaded and converted all data in the test file
+    val source = new DefaultSource(jdbcWrapper)
+    val relation = source.createRelation(testSqlContext, params)
+    val df = testSqlContext.baseRelationToDataFrame(relation)
+
+    df.rdd.collect() zip expectedData foreach {
```
JoshRosen: This is a dangerous way to compare answers since this will never fail if the query returns an empty result:

```scala
scala> (Seq.empty[Int] zip Seq(1, 2, 3)).foreach(_ => assert(false))
```

This will not throw any error.

JoshRosen: I'm going to fix this by porting something similar to SparkSQL's `checkAnswer` function, but for future reference I think that `df.rdd.collect() shouldBe expectedData` would have been simpler and more robust.

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -0,0 +1,486 @@
...
+            EqualTo("testString", "Unicode是樂趣"),
+            GreaterThan("testDouble", 1000.0),
+            LessThan("testDouble", Double.MaxValue),
+            GreaterThanOrEqual("testFloat", 1.0f),
+            LessThanOrEqual("testInt", 43))
+    val rdd = relation.asInstanceOf[PrunedFilteredScan].buildScan(Array("testByte", "testBool"), filters)
+
+    // We should now only have one matching row, with two columns
+    val filteredExpectedValues = Array(Row(1, true))
+    rdd.collect() zip filteredExpectedValues foreach {
```
JoshRosen: The `zip` here is funny given that this test only expects to receive one row as a correct answer.

JoshRosen:
Nit: I would have preferred to use Mockito over ScalaMock since I think more people will be familiar with that framework and it's one less tool to have to learn / debug.

src/main/scala/com/databricks/spark/redshift/Conversions.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import scala.util.parsing.combinator.JavaTokenParsers
+
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+
+/**
+ * Simple parser for Redshift's UNLOAD bool syntax
+ */
+private object RedshiftBooleanParser extends JavaTokenParsers {
```
JoshRosen: This seems like a really roundabout way to turn the strings `t` and `f` into `true` and `false`.  Why is this better than a simple `if` statement?

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -0,0 +1,486 @@
...
+   */
+  private var sc: SparkContext = _
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    sc = new TestContext
+  }
+
+  override def afterAll(): Unit = {
+    val temp = new File(tempDir)
```
JoshRosen: Since we already depend on Guava, why not use one of its file helper methods to delete the non-empty test directory?

PR #26, opened 2015-07-23
Fix overwrite mode for nonexistent table

emlyn:
If the write mode is set to `overwrite`, loading will fail if the target table does not already exist, when it tries to rename it to the backup table.
This only tries to create the backup table if the target table already exists, and only drops the backup table if it was created.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -73,7 +73,9 @@ class RedshiftWriter(jdbcWrapper: JDBCWrapper) extends Logging {
       log.info("Existing data will be backed up in: " + backupTable)
 
       action(tempTable)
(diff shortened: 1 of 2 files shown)


JoshRosen:
This sounds like a reasonable fix to me. I'm going to try to write a regression test for this as part of my other PR to add end-to-end tests against a real Redshift cluster.

traviscrawford:
@eduardoramirez and I were just talking about this same issue today. We're patching this into our branch and will let you know if there are any issues.

JoshRosen:
I've added a regression test for this and verified that this fix works, so I'm going to commit this change as part of 554c9f9ef1a099efaa29fa23a20a3298ee62ba5d, with the commit authorship credited to @emlyn. Thanks for fixing this and sorry for the long delay in getting this merged.

emlyn:
Thanks for merging this

PR #27, opened 2015-07-24
Option to enable Avro compression

emlyn:
I've added an option to enable compression on the Avro data loaded into Redshift. I haven't benchmarked it but it does seem that it can have a quite significant impact on performance (a small Spark job with the output loaded into Redshift went from ~5 minutes to ~4 minutes when I enabled snappy compression, and roughly half of that run time was on the Spark job itself).
Ideally setting snappy compression by default would be nice, as it should be available everywhere so it's basically a free performance boost. But enabling it involves modifying the Hadoop configuration, which could possibly interact subtly with other parts of the user's job and lead to hard to diagnose issues down the line.
So I've left it disabled by default, but included it in the examples so that it is easier to find.
I also cleaned up a couple of typos in the README examples while I was modifying it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ val sqlContext = new SQLContext(sc)
 // Get some data from a Redshift table
 val df: DataFrame = sqlContext.read
     .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass")
+    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
     .option("dbtable", "my_table")
     .option("tempdir", "s3://path/for/temp/data")
     .load()
@@ -60,7 +60,7 @@ val df: DataFrame = sqlContext.read
 
 df.write
     .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass")
+    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
     .option("dbtable", "my_table_copy")
     .option("tempdir", "s3://path/for/temp/data")
     .option("avrocompression", "snappy")
@@ -79,7 +79,7 @@ sql_context = SQLContext(sc)
 # Read data from a table
 df = sql_context.read \
     .format("com.databricks.spark.redshift") \
-    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass") \
+    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass") \
     .option("dbtable", "my_table") \
     .option("tempdir", "s3://path/for/temp/data") \
     .load()
@@ -87,7 +87,7 @@ df = sql_context.read \
 # Write back to a table
 df.write \
     .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass") \
+    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass") \
     .option("dbtable", "my_table_copy") \
     .option("tempdir", "s3://path/for/temp/data") \
     .option("avrocompression", "snappy")
@@ -103,7 +103,7 @@ USING com.databricks.spark.redshift
 OPTIONS (dbtable 'my_table',
          tempdir 's3://my_bucket/tmp',
          avrocompression 'snappy',
-         url 'jdbc:postgresql://host:port/db?user=username&password=pass');
+         url 'jdbc:redshift://host:port/db?user=username&password=pass');
 ```
 
 ### Scala helper functions
--- a/README.md
+++ b/README.md
@@ -279,6 +279,16 @@ It may be useful to have some <tt>GRANT</tt> commands or similar run here when l
 table, the changes will be reverted and the backup table restored if post actions fail.</p>
     </td>
  </tr>
+ <tr>
+    <td><tt>avrocompression</tt></td>
+    <td>No</td>
+    <td><tt>snappy</tt></td>
+    <td>
+<p>Sets the compression codec to use on the Avro data to be loaded into Redshift. This overwrites the <tt>avro.output.codec</tt>
+key in the Hadoop configuration with the specified value. To disable this and use the value set in the Hadoop configuration,
+set this to null or an empty string.</p>
+    </td>
+ </tr>
 </table>
 
 ## AWS Credentials
--- a/README.md
+++ b/README.md
@@ -291,6 +291,13 @@ table, the changes will be reverted and the backup table restored if post action
 key in the Hadoop configuration with the specified value and also sets <tt>mapred.output.compress = true</tt> and
 <tt>mapred.output.compression.type = BLOCK</tt>. If left unset (or set to null or an empty string) it will leave
 the Hadoop configuration unchanged.</p>
+<p>Valid settings are:</p>
+<ul>
+    <li><tt>""</tt> (default): use compression settings from Hadoop config (usually none unless explicitly set).</li>
+    <li><tt>"snappy"</tt>: use snappy compression.</li>
+    <li><tt>"deflate"</tt>: use deflate (zlib) compression (better ratio but more CPU intensive than snappy).</li>
+    <li><tt>"null"</tt>: disable compression.</li>
+</ul>
     </td>
  </tr>
 </table>
--- a/README.md
+++ b/README.md
@@ -288,7 +288,8 @@ table, the changes will be reverted and the backup table restored if post action
     <td>No compression (unless set in Hadoop config)</td>
     <td>
 <p>Sets the compression codec to use on the Avro data to be loaded into Redshift. This overwrites the <tt>avro.output.codec</tt>
-key in the Hadoop configuration with the specified value. If left unset (or set to null or an empty string) it will leave
+key in the Hadoop configuration with the specified value and also sets <tt>mapred.output.compress = true</tt> and
+<tt>mapred.output.compression.type = BLOCK</tt>. If left unset (or set to null or an empty string) it will leave
 the Hadoop configuration unchanged.</p>
     </td>
  </tr>
--- a/README.md
+++ b/README.md
@@ -50,21 +50,22 @@ val sqlContext = new SQLContext(sc)
 // Get some data from a Redshift table
 val df: DataFrame = sqlContext.read
     .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
-    .option("dbtable" -> "my_table")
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/redshift/Parameters.scala
```diff
@@ -194,5 +195,12 @@ private [redshift] object Parameters extends Logging {
 
       s"aws_access_key_id=$accessKeyId;aws_secret_access_key=$secretAccessKey"
     }
+
+    /**
+     * When nonempty/non-null sets the compression codec to use for writing Avro data.
+     *
+     * Defaults to snappy.
```
jaley: Default is actually disabled now, looks like this doc comment needs an update if we're keeping it disabled by default.

emlyn: Thanks, updated.

README.md
```diff
@@ -272,6 +275,17 @@ It may be useful to have some <tt>GRANT</tt> commands or similar run here when l
 table, the changes will be reverted and the backup table restored if post actions fail.</p>
     </td>
  </tr>
+ <tr>
+    <td><tt>avrocompression</tt></td>
+    <td>No</td>
+    <td>No compression (unless set in Hadoop config)</td>
+    <td>
+<p>Sets the compression codec to use on the Avro data to be loaded into Redshift. This overwrites the <tt>avro.output.codec</tt>
```
jaley: Might be nice to enumerate the available options here, to save people researching how it's plumbed through. I believe `"snappy"`, `"bzip2`and `"gzip"` will all work.

Maybe it's worth us supporting an explicit `"none"` option for the sake of clarity? Now that I think about it, the empty string / null option to disable compression is a bit unnecessarily subtle?

emlyn: I'll add the options to the table, I think another valid one is `"deflate"`.
I'm not sure that an empty string is so bad - leaving compression settings alone is not really the same as `none`, unless `none` would explicitly disable compression.

emlyn: According to this: https://avro.apache.org/docs/1.7.7/spec.html#Object+Container+Files only `"deflate"` and `"snappy"` are available (and `"null"` for no compression).

emlyn:
I just noticed this in the output:
```
INFO Configuration.deprecation: mapred.output.compress is deprecated. Instead, use mapreduce.output.fileoutputformat.compress
INFO Configuration.deprecation: mapred.output.compression.type is deprecated. Instead, use mapreduce.output.fileoutputformat.compress.type
```
Should this be setting the hadoop2 properties? Or could it be used with hadoop1 where they're not available?

JoshRosen:
Hey @emlyn, I'm going to be spending a lot more time working on `spark-redshift` this week and was wondering whether you'd be interested in bringing this patch up to date.  Are you blocked awaiting an answer on the Hadoop configuration issue?  AFAIK Spark itself may be setting some of those legacy properties in order to maintain compatibility with Hadoop 1.x.

emlyn:
I've brought this up to date with `master` (sorry, I rebased instead of merging, not quite thinking straight after being away on holiday for a few weeks...).
In order to enable compression, I'm setting the legacy Hadoop properties ([here](https://github.com/emlyn/spark-redshift/blob/compression/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala#L185)), but I'm not sure whether I should follow the deprecation warning advice and use the new properties instead, or use the legacy properties for backwards compatibility (or set both?).

JoshRosen:
Don't worry about rebase vs. merge or the cleanliness of the commit log; I'm going to use a merge script to squash everything into a single commit anyways.

I'll take a closer pass through this later in the afternoon. One high-level comment, though: I think that we'll need to add end-to-end tests for this as part of `RedshiftIntegrationSuite` to make sure that it actually works against a real Redshift cluster. For security reasons, the Travis credential injection won't actually work on pull request builds, so what I'll do is pull your changes into a temporary branch in the `databricks` repo, add the test myself, and check whether it passes.

README.md
```diff
@@ -50,21 +50,22 @@ val sqlContext = new SQLContext(sc)
 // Get some data from a Redshift table
 val df: DataFrame = sqlContext.read
     .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
-    .option("dbtable" -> "my_table")
-    .option("tempdir" -> "s3://path/for/temp/data")
+    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass")
```
JoshRosen: This should be `redshift`. In an earlier commit, I updated the guide to illustrate use of Amazon's official JDBC connector.

README.md
```diff
@@ -50,21 +50,22 @@ val sqlContext = new SQLContext(sc)
...
 
 df.write
-  .format("com.databricks.spark.redshift")
-    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
-    .option("dbtable" -> "my_table_copy")
-    .option("tempdir" -> "s3://path/for/temp/data")
-  .mode("error")
-  .save()
+    .format("com.databricks.spark.redshift")
+    .option("url", "jdbc:postgresql://redshifthost:5439/database?user=username&password=pass")
```
JoshRosen: Same here: `redshift://`

README.md
JoshRosen: We should probably verify that Redshift supports all of these compression modes. I don't see why it would necessarily support fewer of them, but it's possible. I may be able to take care of this as part of integration testing (I might add a new suite which tests all of the compression options).

JoshRosen:
I've opened the https://github.com/databricks/spark-redshift/compare/avro-compression branch to try to write some integration tests for this.

I'm actually not sure whether we want to update the examples to include this; I think that mentioning it in the docs should be sufficient. If we think that's not going to draw enough attention to this feature then I think we should add a separate "performance tips" heading, since other patches will add similar tuning options.

README.md
```diff
@@ -279,6 +282,24 @@ It may be useful to have some <tt>GRANT</tt> commands or similar run here when l
...
+<p>Sets the compression codec to use on the Avro data to be loaded into Redshift. This overwrites the <tt>avro.output.codec</tt>
+key in the Hadoop configuration with the specified value and also sets <tt>mapred.output.compress = true</tt> and
+<tt>mapred.output.compression.type = BLOCK</tt>. If left unset (or set to null or an empty string) it will leave
+the Hadoop configuration unchanged.</p>
+<p>Valid settings are:</p>
+<ul>
+    <li><tt>""</tt> (default): use compression settings from Hadoop config (usually none unless explicitly set).</li>
+    <li><tt>"snappy"</tt>: use snappy compression.</li>
+    <li><tt>"deflate"</tt>: use deflate (zlib) compression (better ratio but more CPU intensive than snappy).</li>
+    <li><tt>"null"</tt>: disable compression.</li>
```
JoshRosen: Is this true? It looks like this just leaves the default configuration options unchanged.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -180,6 +180,12 @@ class RedshiftWriter(jdbcWrapper: JDBCWrapper) extends Logging {
   def saveToRedshift(sqlContext: SQLContext, data: DataFrame, params: MergedParameters) : Unit = {
     val conn = jdbcWrapper.getConnector(params.jdbcDriver, params.jdbcUrl, new Properties()).apply()
 
+    if (params.avrocompression != null && params.avrocompression.nonEmpty) {
+      val conf = sqlContext.sparkContext.hadoopConfiguration
```
JoshRosen: Instead of mutating the (effectively) global Hadoop configuration, I think that it would be better to clone this configuration and pass our own local copy throughout the Redshift library.  I'm going to make a separate pull request to perform this cleanup.

JoshRosen: Actually, there's nowhere else that we mutate this configuration, so this change will have to be part of this patch.

README.md
emlyn: It would overwrite whatever value was set before with the string `"null"`, which I believe is a dummy codec that does no compression (at least I think I read that somewhere, but can't find it now).

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
emlyn: Yes, it wasn't immediately obvious to me how to copy and pass through a new config, but that would be much nicer than mutating the global config.

emlyn:
If we are able to enable compression without mutating the global config, I think it might be nice to enable some form of compression by default, as a fast codec like snappy should be a net gain in all cases.

JoshRosen:
I'd like to merge this feature in but there are some other bugfixes that are taking higher priority for me right now. If you have time to work on this, would you be interested in writing a new integration test suite for this? I'd be happy to walk you through how to run the integration tests locally (you'll need to create a new S3 bucket and Redshift database for this, though).

emlyn:
I agree it would be nice to get this merged, but I'm not sure that we're going to have too much time over the next few weeks. I might be able to spend a bit of time on it if it's not too much work - what do you envisage by "writing a new integration test suite"? Is it just testing that importing into Redshift also works when compression is enabled? We already have a dev Redshift cluster that I presume we could use for this, so hopefully setting it up shouldn't be too hard.

JoshRosen:
What I meant was writing some end-to-end tests which exercise this feature against a real Redshift cluster, similar to what we have in `RedshiftIntegrationSuite` (although I'd add a new suite for compression-related tests, similar to what I did with `STSIntegrationSuite`. Given how many bugs I've found in even fairly simple features, I'm fairly hesitant to merge new features which we don't have tests for, especially if it's not _that_ hard to add them (I'd say an hour, tops, for this feature).

JoshRosen:
I'd also be interested in seeing a benchmarking tool / suite / notebook for this, esp. since it would come in handy for #73.

emlyn:
Is this the sort of thing you mean? I can try to run it locally if you give me some pointers on how to run the integration tests.

JoshRosen:
Yep, that's along the lines of what I have in mind. I'm out for the rest of today but I'll try to take a closer look at this tomorrow or Friday.

JoshRosen:
It looks like Snappy compression is enabled by default in the newer version of `spark-avro` that this library uses: https://github.com/databricks/spark-avro/blame/01ae707bf3dbe3e74cfc8b04e892eb34defd28fe/src/main/scala/com/databricks/spark/avro/AvroRelation.scala#L89

Given that it looks like we've been using compression this whole time starting with 0.5.0+, is this PR still relevant?

emlyn:
In that case, I agree that this is probably no longer needed.

JoshRosen:
Great! I'm going to close this as "wontfix" then.

jomach:
@JoshRosen Any updates on this one ? Thanks

PR #31, opened 2015-07-30
remove errant brace from scheme prefix

koeninger:


JoshRosen:
This is another example of end-to-end test coverage issues.

JoshRosen:
This also looks like a straightforward fix, so I'm going to merge it now.

PR #36, opened 2015-07-31
Bug fixes and auto testing of Redshift data source

kai-zeng:
Major changes:
(1) Configuring aws access key pairs: The previous implementation is not consistent on the aws access key pairs used in unloading and reading data. This PR fixes this bug. 
Now the aws access key pair is configured in this priority order:
<1> If the 'tempdir' option encodes the aws access key pair. That key pair is used.
<2> Otherwise, we use the aws key pair specified in the hadoop configuration of the spark context.

(2) Automating the test. We set up a redshift server for this test, and encode the security information using Travis.

(3) Auto convert column names to lower cases as Redshift COPY only supports lowercase column names.

(4) Other miscellaneous fixes.


kai-zeng:
cc @marmbrus @yhuai

JoshRosen:
I'm going to be taking over this PR.  As it stands now, I think that this PR is too big because it contains both bugfixes and general test coverage improvements. I'm going to see about splitting these into a series of smaller incremental PRs.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
+/*
+ * Copyright 2015 TouchType Ltd
```
JoshRosen: If this is a new file with original code then it should have the DB copyright.

.travis.yml
```diff
@@ -1,17 +1,20 @@
 language: scala
 scala:
-  - 2.10.4
```
JoshRosen: These unrelated whitespace changes are annoying.

README.md
```diff
@@ -180,23 +180,13 @@ need to be configured to allow access from your driver application.
...
-    <td><a href="http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/credentials.html">Default Provider Chain</a></td>
-    <td>AWS secret access key corresponding to provided access key.</td>
- </tr>
- <tr>
     <td><tt>tempdir</tt></td>
     <td>Yes</td>
     <td>No default</td>
     <td>A writeable location in Amazon S3, to be used for unloaded data when reading and Avro data to be loaded into
-Redshift when writing. If you're using `spark-redshift` as part of a regular ETL pipeline, it can be useful to
+Redshift when writing. The location url could contain aws access key id and aws secrete access key. Otherwise, the default
```
JoshRosen: Style: AWS should always be capitalized.

JoshRosen: Also: `secrete`; proofreading.

src/main/scala/com/databricks/spark/redshift/Conversions.scala
```diff
@@ -52,12 +51,12 @@ private [redshift] object Conversions {
   // apparently not clues about this in the table schema. This format delegates to one of the above
   // formats based on string length.
   val redshiftTimestampFormat = new DateFormat() {
-    override def format(date: Date, toAppendTo: StringBuffer, fieldPosition: FieldPosition): StringBuffer = {
+    override def format(date: utilDate, toAppendTo: StringBuffer, fieldPosition: FieldPosition): StringBuffer = {
```
JoshRosen: Type names should generally be capitalized.

src/test/scala/com/databricks/spark/redshift/TestUtils.scala
```diff
@@ -39,17 +39,18 @@ object TestUtils {
    */
   lazy val testSchema =
     StructType(
-      Seq(
```
JoshRosen: I'd back out these unnecessary changes.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+import org.apache.spark.sql.jdbc.DefaultJDBCWrapper
+import org.apache.spark.sql.{Row, SQLContext, SaveMode}
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.{BeforeAndAfterAll, FunSuite, Matchers}
+
+/**
+ * Tests main DataFrame loading and writing functionality
+ */
+class RedshiftSuite extends FunSuite with Matchers with MockFactory with BeforeAndAfterAll {
+  val jdbcUrl = s"${System.getenv("aws_redshift_jdbc_url")}?" +
```
JoshRosen: This is messy / hard-to-read. This could be simplified by just using the `vals` defined two lines down.

Also, I think it's generally conventional to put environment variables in all-caps.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+import java.util.Properties
+
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.jdbc.DefaultJDBCWrapper
+import org.apache.spark.sql.{Row, SQLContext, SaveMode}
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.{BeforeAndAfterAll, FunSuite, Matchers}
+
+/**
+ * Tests main DataFrame loading and writing functionality
```
JoshRosen: This comment is too short to describe this test suite. I would have expanded this to explain that this runs an end-to-end integration test against a real Redshift cluster.

I'd also consider moving this out of the default unit test lifecycle and into an integration test lifecycle so that this test may be more easily skipped outside of Travis.

JoshRosen:
> (1) Configuring aws access key pairs: The previous implementation is not consistent on the aws access key pairs used in unloading and reading data. This PR fixes this bug. 
Now the aws access key pair is configured in this priority order:
> <1> If the 'tempdir' option encodes the aws access key pair. That key pair is used.
> <2> Otherwise, we use the aws key pair specified in the hadoop configuration of the spark context.

Would have been good to compare / contrast this with the fix proposed by @koeninger in #32.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -38,14 +38,11 @@ case class RedshiftRelation(jdbcWrapper: JDBCWrapper, params: MergedParameters,
...
 
-  override def schema = {
-    userSchema match {
-      case Some(schema) => schema
-      case None => {
-        jdbcWrapper.registerDriver(params.jdbcDriver)
-        jdbcWrapper.resolveTable(params.jdbcUrl, params.table, new Properties())
-      }
-    }
+  override def schema = userSchema match {
```
JoshRosen: If you're going to reformat this anyways then you might as well replace this with a `.getOrElse`.

JoshRosen:
This PR is _really_ difficult to review / understand in its current state, so I'm going to open a new WIP PR where I try to gradually revert some of the unnecessary changes in order to figure out which fixes were actually necessary.

src/test/scala/com/databricks/spark/redshift/ParametersSuite.scala
```diff
@@ -67,4 +69,34 @@ class ParametersSuite extends FunSuite with Matchers {
...
+
+  test("invalid options with both dbtable and query specified") {
+    val params =
+      Map(
+        "tempdir" -> "s3://keyId1:secretKey1@foo/bar",
+        "dbtable" -> "test_table",
+        "query" -> "select * from test_table",
+        "url" -> "jdbc:postgresql://foo/bar")
+
+    intercept[Exception] {
```
JoshRosen: Is there a more specific exception that we could expect?

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -96,12 +108,12 @@ case class RedshiftRelation(jdbcWrapper: JDBCWrapper, params: MergedParameters,
   }
 
   protected def compileValue(value: Any): Any = value match {
-    case stringValue: String => s"\\'${escapeSql(stringValue.toString)}\\'"
+    case stringValue: UTF8String => s"\\'${escapeSql(stringValue.toString)}\\'"
```
JoshRosen: Why are we using `UTF8String` here, given that it's not intended to be a public API?

build.sbt
```diff
@@ -10,7 +10,7 @@ version := "0.4.1-SNAPSHOT"
 
 scalaVersion := "2.10.4"
 
-sparkVersion := sys.props.get("spark.version").getOrElse("1.4.1")
+sparkVersion := sys.props.get("spark.version").getOrElse("1.5.0-SNAPSHOT")
```
JoshRosen: Why change the default Spark version to 1.5 unless we're dropping support for 1.4.x?

JoshRosen:
Closing this in favor of #41.

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -231,20 +233,50 @@ class RedshiftSourceSuite
...
+        GreaterThan("testDouble", 1234152.123124981),
+        LessThan("testDouble", Double.MaxValue),
+        GreaterThanOrEqual("testFloat", 1.0f),
+        LessThanOrEqual("testInt", 42))
+    val rdd = relation.asInstanceOf[PrunedFilteredScan].buildScan(Array("testByte", "testBool"), filters)
+
+    // We should now only have one matching row, with two columns
+    val filteredExpectedValues = Array(Row(1, true))
+    rdd.collect() zip filteredExpectedValues foreach {
+      case (loaded, expected) => loaded shouldBe expected
```
JoshRosen: Nit: I don't like should matchers. Also, it would be nice to use something similar to Spark SQL's `checkAnswer` so that you get useful debugging output for failures.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+
+    createTable("test_table")
+    createTable("test_table2")
+    createTable("test_table3")
+
+    conn.commit()
+  }
+
+  override def afterAll(): Unit = {
+    val temp = new File(tempDir)
```
JoshRosen: Isn't `tempDir` an `s3://...` URI? If so, this won't be able to delete files on S3.

I think that we should be using a Hadoop filesystem instead.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+
+    sc.stop()
+    super.afterAll()
+  }
+  
+  test("DefaultSource can load Redshift UNLOAD output to a DataFrame") {
+    val sqlContext = new SQLContext(sc)
+    sqlContext.sql(
+      s"""
+         |create temporary table test_table(
```
JoshRosen: It looks like every test case creates its own temporary table on a fresh SQL context.  I think that it would make more sense to extract this out into a `beforeEach` method, since the code duplication here makes the tests fairly hard to read.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    sc = new SparkContext("local", "RedshiftSourceSuite")
+
+    sc.hadoopConfiguration.set("fs.s3n.awsAccessKeyId", accessKeyId)
+    sc.hadoopConfiguration.set("fs.s3n.awsSecretAccessKey", secretAccessKey)
+
+    conn = DefaultJDBCWrapper.getConnector("org.postgresql.Driver", jdbcUrl, new Properties())()
+
+    conn.prepareStatement("drop table if exists test_table").executeUpdate()
```
JoshRosen: If there are multiple parallel Travis builds sharing the same database + redshift instance then they may clobber each other due to the common table names.  To make the tests more robust, I think that we should append a random suffix to our table names.

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+           |testint int4,
+           |testlong int8,
+           |testshort int2,
+           |teststring varchar(256),
+           |testtimestamp timestamp
+           |)
+      """.stripMargin
+      ).executeUpdate()
+
+      conn.prepareStatement(
```
JoshRosen: Why create a separate prepared statement for each row instead of creating a single parameterized prepared statement?

JoshRosen: Actually, why not just create a single big multi-row insert statement?

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
JoshRosen: Actually, I see now that most of this suite is based on / copied from RedshiftSourceSuite.  It would have been good to leave a comment explaining this (maybe instead of the uninformative comment that's currently at the top of this suite). It would have also been nice to factor out some of the code into common helper classes (e.g. not repeating `expectedData` in both suites).

src/test/scala/com/databricks/spark/redshift/RedshiftSuite.scala
```diff
@@ -0,0 +1,458 @@
...
+      Row(1.toByte, true)
+    )
+
+    sqlContext.sql("select testbyte, testbool from test_table order by testbyte, testbool").collect()
+      .zip(prunedExpectedValues).foreach {
+      case (loaded, expected) => loaded shouldBe expected
+    }
+  }
+
+  test("DefaultSource supports user schema, pruned and filtered scans") {
```
JoshRosen: I think that this test should be renamed in this suite since it's not actually asserting that a pruned or filtered scan is taking place (i.e. it's not directly testing whether pushdown actually took place). If we keep the name, then we might want to strengthen the assertion to actually assert something about the number of rows loaded from Redshift vs. the total number of rows in the DB.

PR #44, opened 2015-08-19
Clean up RedshiftSourceSuite and improve assertions

JoshRosen:
This patch performs cleanup in `RedshiftSourceSuite` and improves a few assertions.

I factored out lots of shared setup code, simplifying many tests.

In addition, I fixed some bad assertions that were added in #22.  The problem with the existing assertions is that they `zip` together collections to compare them element-by-element but do so without first checking that the collections contain the same number of elements. As a result, these assertions can silently pass in cases where Spark returns an empty answer or an answer with more rows than expected.

I've fixed this by porting Spark SQL's `QueryTest` helper and by using whole-collection comparisons instead of element-by-element comparisons (this shouldn't be a problem given that our expected answers only contain a handful of rows).

The assertion fix exposed a problem in one of the filter pushdown unit tests, so I strengthened these tests by adding assertions which inspect the actual `UNLOAD` queries that would be submitted to Redshift.


src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -224,11 +217,7 @@ class RedshiftSourceSuite
             LessThanOrEqual("testInt", 43))
     val rdd = relation.asInstanceOf[PrunedFilteredScan].buildScan(Array("testByte", "testBool"), filters)
 
-    // We should now only have one matching row, with two columns
```
JoshRosen: Surprise! The bad assert masked a bug:

```scala
[info] - DefaultSource supports user schema, pruned and filtered scans *** FAILED ***
[info]   Array([1,true], [1,false], [0,null], [0,false], [null,null]) did not equal Array([1,true]) (RedshiftSourceSuite.scala:220)
```

JoshRosen: Actually, I don't know that this test necessarily makes sense as written: it seems like we want our unit test to simply assert that the filter gets pushed down into Redshift. The end-to-end test to check whether we properly translate filters into `WHERE` clauses probably needs to run against a real database, so we can test the returned rows as part of that test.

For this test, I think that we should just write assertions over the exact queries passed to the mock database connection. I'm going to refactor this suite to do that.

Before I do that, though, a lot of the minor style and formatting issues in these files are bothering me, so I'm going to submit a separate style cleanup PR to implement Scalastyle checks so we never have to deal with this issue again.

JoshRosen:
I've decided to broaden the scope of this PR into a more general refactoring / splitting of this suite in order to improve understandability and conciseness.

JoshRosen:
I found a nice way to fix the assertion: add an assert which inspects the actual SQL that's send to Redshift. This will help to guard against unintentional change to the generated queries.

JoshRosen:
I'm going to commit this now; will address any concerns in followups.

JoshRosen:
This didn't find / fix all of the bugs in the asserts and there are still some masked bugs that need fixing; see #50.

Issue #40, opened 2015-08-18
Reliance on Spark's private JDBCRDD API breaks compilation against Spark 1.5.0


JoshRosen:
This library's use of Spark's private `JDBCRDD` API means that it no longer compiles against Spark 1.5.0:

```scala
/build/sbt -Dspark.version=1.5.0-SNAPSHOT clean test

[error] /Users/joshrosen/Documents/spark-redshift/src/main/scala/org/apache/spark/sql/jdbc/RedshiftJDBCWrapper.scala:33: not found: value JDBCRDD
[error]     JDBCRDD.getConnector(driver, url, properties)
[error]     ^
[error] /Users/joshrosen/Documents/spark-redshift/src/main/scala/org/apache/spark/sql/jdbc/RedshiftJDBCWrapper.scala:34: not found: value JdbcUtils
[error]   def tableExists(conn: Connection, table: String) = JdbcUtils.tableExists(conn, table)
[error]                                                      ^
[error] /Users/joshrosen/Documents/spark-redshift/src/main/scala/org/apache/spark/sql/jdbc/RedshiftJDBCWrapper.scala:29: not found: value DriverRegistry
[error]   def registerDriver(driverClass: String) = DriverRegistry.register(driverClass)
[error]                                             ^
[error] /Users/joshrosen/Documents/spark-redshift/src/main/scala/org/apache/spark/sql/jdbc/RedshiftJDBCWrapper.scala:31: not found: value JDBCRDD
[error]     JDBCRDD.resolveTable(jdbcUrl, table, properties)
[error]     ^
[error] /Users/joshrosen/Documents/spark-redshift/src/main/scala/org/apache/spark/sql/jdbc/RedshiftJDBCWrapper.scala:28: not found: value JDBCWriteDetails
[error]   def schemaString(dataFrame: DataFrame, url: String) = JDBCWriteDetails.schemaString(dataFrame, url)
[error]                                                         ^
[error] 5 errors found
[error] (compile:compile) Compilation failed
[error] Total time: 21 s, completed Aug 18, 2015 9:23:39 AM
```

The problem is that these class's package changed from `org.apache.spark.sql.jdbc` to `org.apache.spark.sql.execution.datasources.jdbc`.

[...]

PR #46, opened 2015-08-19
Copy Spark code to minimize reliance on private APIs

JoshRosen:
This commit copies JDBC code from Spark in order to minimize this library's reliance on private Spark APIs. This partially addresses #40; that issue can be completely resolved once we publish Maven snapshots for 1.5.0 and are able to test against that version in Travis.


src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -0,0 +1,210 @@
...
+
+import org.apache.spark.Logging
+import org.apache.spark.sql.types._
+
+/**
+ * Shim which exposes some JDBC helper functions.
+ */
+private[redshift] class JDBCWrapper extends Logging {
+
+  def registerDriver(driverClass: String): Unit = {
```
marmbrus: Can you add a comment about what is going on here and why we are using reflection.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -0,0 +1,210 @@
...
+      case java.sql.Types.SMALLINT      => IntegerType
+      case java.sql.Types.SQLXML        => StringType
+      case java.sql.Types.STRUCT        => StringType
+      case java.sql.Types.TIME          => TimestampType
+      case java.sql.Types.TIMESTAMP     => TimestampType
+      case java.sql.Types.TINYINT       => IntegerType
+      case java.sql.Types.VARBINARY     => BinaryType
+      case java.sql.Types.VARCHAR       => StringType
+      case _                            => null
+      // scalastyle:on
```
marmbrus: maybe a TODO to clean up the ones that are irrelevant for redshift.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -0,0 +1,210 @@
...
+import java.sql.{Connection, DriverManager, ResultSetMetaData, SQLException}
+import java.util.Properties
+
+import scala.util.Try
+
+import org.apache.spark.Logging
+import org.apache.spark.sql.types._
+
+/**
+ * Shim which exposes some JDBC helper functions.
```
marmbrus: Maybe mention that this came from Spark SQL so that people aren't confused by extra stuff that redshift doesn't support.

marmbrus:
a few comments but generally LGTM

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -0,0 +1,210 @@
...
+    schema.fields.foreach { field => {
+      val name = field.name
+      val typ: String = field.dataType match {
+        case IntegerType => "INTEGER"
+        case LongType => "BIGINT"
+        case DoubleType => "DOUBLE PRECISION"
+        case FloatType => "REAL"
+        case ShortType => "INTEGER"
+        case ByteType => "SMALLINT" // Redshift does not support the BYTE type.
+        case BooleanType => "BIT(1)"
```
JoshRosen: Turns out that Redshift doesn't support this type either.

JoshRosen:
Updated to address comments.  Still need to update to fix a bug which isn't caught by the existing tests (but which is caught in my testing branch). Once I do that, I'm going to merge this.

Issue #39, opened 2015-08-09
DATE columns unloaded as epochmillisecs can not be loaded back into Redshift


gregrahn:
Both DATE and TIMESTAMP types are unloaded using epochmillisecs format currently.  Data is loaded back into Redshift by using `COPY ... TIMEFORMAT 'epochmillisecs'` [here](https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala#L59), however, this fails for DATE types as the default of  `DATEFORMAT 'auto'`, does not handle epochmillisecs, nor is there currently even an option of  `DATEFORMAT 'epochmillisecs'` that could be used (perhaps a good Redshift enhancement).  Given this, it seems to make more sense to use a common serialization format that could be used for both DATE and TIMESTAMP, like `YYYY-MM-DD HH:MI:SS.SSSSSS` which does work for both types even when the data contains fractional seconds using `DATEFORMAT 'YYYY-MM-DD HH:MI:SS'`.  The default format for TIMESTAMP is  `YYYY-MM-DD HH:MI:SS` so no `TIMEFORMAT` is then required. See [docs](http://docs.aws.amazon.com/redshift/latest/dg/r_DATEFORMAT_and_TIMEFORMAT_strings.html).

JoshRosen:
I just ran into a similar issue while trying to improve this package's test coverage; some of our round-trip tests are broken because of this (this problem existed in old code but was masked by bad test assertions: #44).

JoshRosen:
I'll try to fix this tomorrow / this weekend, but pull requests are also welcome if you want to take a stab at it.

JoshRosen:
Alright, I think that I have a fix for this as part of #50.

JoshRosen:
@gregrahn, I've merged #50, which should contain a fix for this issue. There might be some followup work related to timestamp precision, but I think we should file a separate issue to track any problems that arise there. Therefore, I'm going to close this as resolved but please feel free to comment / re-open if you spot any problems with my fix.

PR #50, opened 2015-08-21
Fix problem with QueryTest asserts and fix several masked bugs

JoshRosen:
The version of `QueryTest.checkAnswer` that I ported didn't actually contain any asserts or `fail` calls, so some bugs were silently being masked:

- The tests which read unloaded Avro back need to account for the fact that the Avro files are written to a random subdirectory of `tempdir`.
- Due to a Spark bug where UTF8Strings could be exposed in data source Filter expressions, literals were not being properly escaped in generated SQL, causing them to be treated as identifiers and leading to confusing Unicode errors.  For instance, the following invalid query was being generated:

   ```sql
   ('SELECT "testbool", "teststring", "testint", "testbyte", "testfloat", "testdouble"
   FROM test_table_7620331297344054579
   WHERE "testbool" = true AND "teststring" = Unicode's樂趣 AND "testdouble" = 1234152.12312498 AND "testint" = 42')
   ```
- After fixing the UTF8String issue, there was another issue with escaping of single quotes: `compileValue` escapes the single quotes that it wraps around the string literal, which means that single quotes appearing inside of the string literal also need to be escaped in `escapeSql`.
- Dates and timestamps were encoded in a format that could not be round-tripped to/from Redshift; see #39.
- The test data had too much precision for certain double values. Some Unicode strings were inconsistent between test data and test assertions.
- Timestamp values were being used to represent date when reading the UNLOAD output in Spark; this led to internal type mismatch issues because Spark expects DateType columns to have `java.sql.Date` objects as their public types. This error was difficult to diagnose because of a bug in Spark 1.4.x's implementation of CatalystTypeConverters where unexpected input types would be silently passed through without conversion (in 1.5.x, this raises an informative error message).


JoshRosen:
The following unit tests fail due to empty answers being returned:

```
[info] - DefaultSource serializes data as Avro, then sends Redshift COPY command *** FAILED ***
[info] - Append SaveMode doesn't destroy existing data *** FAILED ***
```

In addition, the following end-to-end integration tests fail with runtime errors:

```
[info] - query with pruned and filtered scans *** FAILED ***
```

fails due to some sort of ASCII character handling error.  It looks like all of the saving tests are failing due to some Timestamp casting error:

```
[info] - DefaultSource can load Redshift UNLOAD output to a DataFrame *** FAILED ***
[info] - Append SaveMode doesn't destroy existing data *** FAILED ***
[info] - Do nothing when table exists if SaveMode = Ignore *** FAILED ***
```

```
[info]   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 2, localhost): java.lang.ClassCastException: java.sql.Timestamp cannot be cast to java.lang.Integer
[info]   	at scala.runtime.BoxesRunTime.unboxToInt(BoxesRunTime.java:106)
[info]   	at org.apache.spark.sql.execution.SparkSqlSerializer2$$anonfun$createSerializationFunction$1.apply(SparkSqlSerializer2.scala:314)
[info]   	at org.apache.spark.sql.execution.SparkSqlSerializer2$$anonfun$createSerializationFunction$1.apply(SparkSqlSerializer2.scala:227)
[info]   	at org.apache.spark.sql.execution.Serializer2SerializationStream.writeKey(SparkSqlSerializer2.scala:65)
[info]   	at org.apache.spark.storage.DiskBlockObjectWriter.write(BlockObjectWriter.scala:206)
[info]   	at org.apache.spark.util.collection.WritablePartitionedIterator$$anon$3.writeNext(WritablePartitionedPairCollection.scala:104)
[info]   	at org.apache.spark.util.collection.ExternalSorter.spillToPartitionFiles(ExternalSorter.scala:375)
[info]   	at org.apache.spark.util.collection.ExternalSorter.insertAll(ExternalSorter.scala:208)
[info]   	at org.apache.spark.shuffle.sort.SortShuffleWriter.write(SortShuffleWriter.scala:62)
[info]   	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:70)
[info]   	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:41)
[info]   	at org.apache.spark.scheduler.Task.run(Task.scala:70)
[info]   	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:213)
[info]   	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[info]   	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
[info]   	at java.lang.Thread.run(Thread.java:745)
[info]
```

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -98,12 +98,13 @@ private[redshift] case class RedshiftRelation(
   }
 
   protected def compileValue(value: Any): Any = value match {
-    case stringValue: String => s"\\'${escapeSql(stringValue.toString)}\\'"
-    case _ => value
+    case stringValue: String => s"\\'${escapeSql(stringValue)}\\'"
+    case stringValue: UTF8String => s"\\'${escapeSql(stringValue.toString())}\\'"
```
JoshRosen: This is a Spark bug, since we shouldn't ever be exposing `UTF8String` to third-party code. We can avoid directly relying on UTF8String if we use the schema to look up column types instead. I'm going to perform this refactoring now.

JoshRosen:
Tests are now passing so this should be ready for review.

src/test/scala/com/databricks/spark/redshift/QueryTest.scala
```diff
@@ -19,20 +19,18 @@ package com.databricks.spark.redshift
...
 import org.apache.spark.sql.catalyst.plans.logical
 import org.apache.spark.sql.{Row, DataFrame}
+import org.scalatest.FunSuite
 
 /**
- * Copy of Spark SQL's `QueryTest` object.
+ * Copy of Spark SQL's `QueryTest` trait.
  */
-object QueryTest {
+trait QueryTest extends FunSuite {
```
marmbrus: +1 to this change.  I much prefer this style to having companion objects, since invocation is less verbose and its less error prone to let checkAnswer actually assert the check.  I'd like to change Spark back as well.

src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
```diff
@@ -263,21 +263,30 @@ class RedshiftIntegrationSuite
...
-          | and testdouble = 1234152.12312498
-          | and testfloat = 1.0
-          | and testint = 42
-        """.stripMargin),
-      Seq(Row(1, true)))
-    // scalastyle:on
+  test("roundtrip save and load") {
+    val tableName = s"roundtrip_save_and_load_$randomSuffix"
+    try {
+      sqlContext.createDataFrame(sc.parallelize(TestUtils.expectedData), TestUtils.testSchema)
```
marmbrus: Doesn't have to be in this PR, but in general I think I prefer tests that declare their own data, using  `Seq[TupleN].toDF()`.  This is concise, and keeps the data near the test so you can spot problems without needing to cross reference.

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -155,9 +155,9 @@ class RedshiftSourceSuite
 
   test("DefaultSource can load Redshift UNLOAD output to a DataFrame") {
     val expectedQuery = (
-      "UNLOAD \\('SELECT \"testByte\", \"testBool\", \"testDate\", \"testDouble\"," +
-      " \"testFloat\", \"testInt\", \"testLong\", \"testShort\", \"testString\", " +
-      "\"testTimestamp\" " +
+      "UNLOAD \\('SELECT \"testbyte\", \"testbool\", \"testdate\", \"testdouble\"," +
```
marmbrus: Nit: I'd prefer `"""` given the amount of escaping needed here.

marmbrus:
this LGTM, comments can be addressed in a followup.

JoshRosen:
Thanks for the review. I'm going to merge this now and will address followups in subsequent PRs.

PR #51, opened 2015-08-21
Fix roundtrip save/load when column names contain uppercase characters

JoshRosen:
Our current write implementation faces multiple issues related to differences in how Redshift and Avro handle case-sensitive column names.

#### Issue 1: columns with uppercase characters in their names cannot be loaded into Redshift tables

In Redshift, [field names are case-insensitive and are folded to lower case](http://docs.aws.amazon.com/redshift/latest/dg/r_names.html).

In Avro, on the other hand, [field names are case-sensitive](http://avro.apache.org/docs/1.7.5/spec.html#Names).

This causes problems when trying to do a roundtrip save and load of a table whose schema contains columns whose names contain uppercase characters. It seems that these columns are silently skipped during Redshift loading, leading to unexpected `null` values for those columns (see the regression test added in this patch for an example). Redshift's docs on [COPYing data from Avro format](http://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-avro) explain why this happens: (emphasis mine):

> COPY automatically maps the data elements in the Avro source data to the columns in the target table by matching field names in the Avro schema to column names in the target table. **Column names are always lowercase, so matching field names must also be lowercase.** With the default 'auto' argument, COPY recognizes only the first level of fields, or outer fields, in the structure.
>
> By default, COPY attempts to match all columns in the target table to Avro field names. You can optionally specify a column list.
>
> [...]
>
> If a column is included in the column list and COPY does not find a matching field in the Avro data, then **COPY attempts to load NULL to the column.**

In order to fix this issue, I propose that `spark-redshift` automatically converts column names to lowercase before saving data to Avro.

#### Issue 2: schemas with columns whose names vary only by capitalization/case should not be used to create Redshift tables

The analyzer should fail fast with an informative error message if a user attempts to write a Redshift table using a schema containing columns with names that vary only by capitalization (e.g. `fooBar` and `foobar`).

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -156,18 +156,27 @@ class RedshiftWriter(jdbcWrapper: JDBCWrapper) extends Logging {
       Row.fromSeq(convertedValues)
     }
 
+    // Convert all column names to lowercase, which is necessary for Redshift to be able to load
+    // those columns (see #51).
+    val schemaWithLowercaseColumnNames: StructType =
+      StructType(data.schema.map(f => f.copy(name = f.name.toLowerCase)))
+
+    if (schemaWithLowercaseColumnNames.map(_.name).toSet.size != data.schema.size) {
+      throw new IllegalArgumentException(
+        "Cannot save table to Redshift because two or more column names would be identical" +
+        " after conversion to lowercase: " + data.schema.map(_.name))
+    }
+
     // Update the schema so that Avro writes date and timestamp columns as formatted timestamp
     // strings. This is necessary for Redshift to be able to load these columns (see #39).
-    // In addition, convert all column names to lowercase, which is necessary for Redshift to be
-    // able to load those columns (see #51).
     val convertedSchema: StructType = StructType(
-      data.schema.map {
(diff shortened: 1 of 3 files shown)


JoshRosen:
Whoops, let me fix the conflict...

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -158,15 +158,28 @@ class RedshiftWriter(jdbcWrapper: JDBCWrapper) extends Logging {
...
-    // Update the schema so that Avro writes these columns as strings:
+    // Convert all column names to lowercase, which is necessary for Redshift to be able to load
+    // those columns (see #51).
+    val schemaWithLowercaseColumnNames: StructType =
+      StructType(data.schema.map(f => f.copy(name = f.name.toLowerCase)))
+
+    if (schemaWithLowercaseColumnNames.map(_.name).toSet.size != data.schema.size) {
+      throw new IllegalArgumentException(
+        "Cannot save table to Redshift because two or more column names would be identical" +
+        " after conversion to lowercase: " + data.schema.map(_.name))
```
marmbrus: `.mkString(", ")`

marmbrus:
LGTM

PR #52, opened 2015-08-25
Allow the results of Redshift queries to be loaded into DataFrames

### LINKED ISSUE #24: Handle full queries as well as table names.
We should be able to handle full queries that use things like redshift joins or aggregation.  In Spark JDBC we do this by wrapping the query in a subquery.  I propose the following interface.  If `query` is specified instead of `dbtable` lets wrap it in `()` and pass it as a subquery.

JoshRosen:
This patch adds support for creating DataFrames from the results of Redshift queries:

- The `dbname` parameter can now accept a query that is wrapped in parentheses, e.g. `(select count(*) from foo`).
- The new `query` parameter is syntactic sugar which lets users omit those parentheses.
- These options are mutually-exclusive and cannot be specified at the same time.
- Only `dbname` can be used for saving data back to Redshift.

This fixes #24.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ val df: DataFrame = sqlContext.read
     .option("tempdir" -> "s3://path/for/temp/data")
     .load()
 
+// Can also load data from a Redshift query
+val df: DataFrame = sqlContext.read
+    .format("com.databricks.spark.redshift")
+    .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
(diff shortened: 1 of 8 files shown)


JoshRosen:
I should add more unit tests to cover the actual `UNLOAD` queries that get sent to Redshift. I also need to add a test which runs a grouped aggregation query in Redshift. Will update shortly.

marmbrus:
LGTM

PR #53, opened 2015-08-26
Improve error messages for Redshift load errors

JoshRosen:
This patch improves our error reporting for Redshift LOAD errors. When a load error occurs, we will now try to automatically fetch more detailed error information from Redshift's [STL_LOAD_ERRORS](http://docs.aws.amazon.com/redshift/latest/dg/r_STL_LOAD_ERRORS.html ) table.

As an example of the improved error messages:

Old:

```
java.sql.SQLException: [Amazon](500310) Invalid operation: Load into table 'error_message_when_string_too_long_3596907251636891354' failed.  Check 'stl_load_errors' system table for details.;
```

New:

```
java.sql.SQLException: Error #1204 while loading data into Redshift: "String length exceeds DDL length".
Table name: the_table_name
Column name: a
Column type: varchar(256)
Raw line: [...]
Raw field value: [...]
```

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -388,7 +388,6 @@ class RedshiftIntegrationSuite
           .save()
       }
       assert(e.getMessage.contains("while loading data into Redshift"))
-
     } finally {
       conn.prepareStatement(s"drop table if exists $tableName").executeUpdate()
       conn.commit()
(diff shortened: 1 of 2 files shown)


JoshRosen:
Hmm, I guess this should also include the destination table name in the error message.

jaley:
This is awesome, thanks for adding this, much sanity saved here :) 

I don't recall exactly what the default permissions situation is for the load errors table. If it's the case that new users need an explicit grant to be run before they can query it, we might want to add a note to the docs that tells users they need to do this to enable better error messages. 

It's possible that Redshift actually does something magic where you can only read rows added by loads from your own user account, in which case it should just work I guess.

emlyn:
Nice! I believe anyone can read from `stl_load_errors` and Redshift will only return rows that relate to the current user (at least that what I've seen when querying it).

marmbrus:
LGTM

PR #54, opened 2015-08-26
Use maxlength metadata to configure VARCHAR column lengths

JoshRosen:
This patch allows users to specify a `maxlength` column metadata entry for string columns in order to control the width of `VARCHAR` columns in generated Redshift table schemas. This is necessary in order to support string columns that are wider than 256 characters. In addition, this configuration can be used as an optimization to achieve space-savings in Redshift. For more background on the motivation of this feature, see #29.

See also: #53 to improve error reporting when LOAD fails.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -154,7 +154,7 @@ private[redshift] class JDBCWrapper extends Logging {
         case _ => throw new IllegalArgumentException(s"Don't know how to save $field to JDBC")
       }
       val nullable = if (field.nullable) "" else "NOT NULL"
-      sb.append(s", $name $typ $nullable")
+      sb.append(s", $name $typ $nullable".trim)
     }}
     if (sb.length < 2) "" else sb.substring(2)
   }
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -141,7 +141,13 @@ private[redshift] class JDBCWrapper extends Logging {
         case ShortType => "INTEGER"
         case ByteType => "SMALLINT" // Redshift does not support the BYTE type.
         case BooleanType => "BOOLEAN"
-        case StringType => "TEXT"
+        case StringType =>
+          val maxlength = if (field.metadata.contains("maxlength")) {
+            field.metadata.getLong("maxlength")
+          } else {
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -141,7 +141,13 @@ private[redshift] class JDBCWrapper extends Logging {
         case ShortType => "INTEGER"
         case ByteType => "SMALLINT" // Redshift does not support the BYTE type.
         case BooleanType => "BOOLEAN"
-        case StringType => "TEXT"
+        case StringType =>
+          val maxlength = if (field.metadata.contains("maxlength")) {
+            field.metadata.getLong("maxlength")
+          } else {
+            255 // TODO: make this default configurable?
```
JoshRosen: Whoops, this should be 256.

Also, should this be configurable?

JoshRosen:
This is a minimum viable patch to illustrate the basic idea and to gather feedback. I'll add more thorough tests tomorrow.

@marmbrus, one question on naming: throughout `spark-redshift` we've used all-lowercase names for many of the library's configuration parameters. I think that you originally proposed a camel-cased name (`maxLength`). Given that you've also proposed using a similar column metadata option in Spark itself (https://github.com/apache/spark/pull/8374), how do you think we should handle capitalization here? Are column metadata field names case-insensitive, in which case this point is moot?

Another question: in the future, is there ever a case where we'd want to generate `CHAR` columns instead of `VARCHAR`? Will this approach be compatible with that (I think so, but we should confirm)?

@Aerlinger, @jaley, @traviscrawford, and @eduardoramirez, you might also want to follow this PR to make sure that it addresses your respective use-cases.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
marmbrus: Should we just use `TEXT` in this case, which essentially defers the default to redshift?

marmbrus:
I believe metadata is case sensitive (which is unfortunate since options are not).  Lowercase looks good to me.  We should perhaps check what the convention in mllib is as they are the heaviest users of this feature.

JoshRosen:
It turns out that the user-facing APIs for manipulating existing columns' metadata are somewhat incomplete in the Scala API and are missing in the Python, SQL, and R language APIs. I don't think that this is a blocker for merging this patch, however, as I feel that the right approach is to improve Spark's APIs for working with column metadata.

In followup PRs, we can consider whether we want to add additional configurations for changing the default size or enabling truncation for working around errors due to limited column width.

PR #55, opened 2015-08-26
Use "aws_access_key_id" and "aws_secret_access_key" for reads, not just unload

JoshRosen:
This is an updated version of #32 with two additional changes: 

- Do not mutate the SparkContext's `hadoopConfiguration`.
- Refactor tests so that they're capable of catching this bug.


JoshRosen:
Going to merge this now, with the commit authorship credited to @koeninger.

PR #57, opened 2015-08-27
Bugfix and optimizations for scans that select no columns

JoshRosen:
If a query resulted in a Redshift scan that selected no columns (such as calling `.count()` on a DataFrame created from a Redshift table), then `spark-redshift` would generate an invalid `UNLOAD` query.

This patch fixes this issue by converting these Redshift scans into `SELECT count(*)` queries, bypassing the unload. This has the added benefit of offering a large performance improvement for such queries.

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -241,7 +241,25 @@ class RedshiftIntegrationSuite
   }
 
   test("count() on DataFrame created from a Redshift table") {
-    assert(sqlContext.sql("select * from test_table").count() === TestUtils.expectedData.length)
+    checkAnswer(
+      sqlContext.sql("select count(*) from test_table"),
+      Seq(Row(TestUtils.expectedData.length))
+    )
+  }
+
+  test("count() on DataFrame created from a Redshift query") {
+    val loadedDf = sqlContext.read
+      .format("com.databricks.spark.redshift")
+      .option("url", jdbcUrl)
+      // scalastyle:off
+      .option("query", s"select * from $test_table where teststring = 'Unicode''s樂趣'")
+      // scalastyle:on
+      .option("tempdir", tempDir)
+      .load()
+    checkAnswer(
+      loadedDf.selectExpr("count(*)"),
+      Seq(Row(1))
+    )
   }
 
   test("Can load output when 'dbtable' is a subquery wrapped in parentheses") {
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -240,6 +240,10 @@ class RedshiftIntegrationSuite
       TestUtils.expectedData)
   }
 
+  test("count() on DataFrame created from a Redshift table") {
+    assert(sqlContext.sql("select * from test_table").count() === TestUtils.expectedData.length)
+  }
+
   test("Can load output when 'dbtable' is a subquery wrapped in parentheses") {
     // scalastyle:off
     val query =
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -99,7 +99,7 @@ private[redshift] case class RedshiftRelation(
         classOf[java.lang.Long], classOf[Array[String]], hadoopConf)
       val prunedSchema = pruneSchema(schema, requiredColumns)
       rdd.values.mapPartitions { iter =>
-        val converter: Array[String] => Row = Conversions.rowConverter(prunedSchema)
+        val converter: Array[String] => Row = Conversions.createRowConverter(prunedSchema)
         iter.map(converter)
       }
     }
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -49,48 +49,72 @@ private[redshift] case class RedshiftRelation(
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -49,50 +49,74 @@ private[redshift] case class RedshiftRelation(
...
+      }
+      // Create a DataFrame to read the unloaded data:
+      val sc = sqlContext.sparkContext
+      val hadoopConf = new Configuration(sc.hadoopConfiguration)
+      params.setCredentials(hadoopConf)
+      val rdd = sc.newAPIHadoopFile(params.tempPath, classOf[RedshiftInputFormat],
+        classOf[java.lang.Long], classOf[Array[String]], hadoopConf)
+      val prunedSchema = pruneSchema(schema, requiredColumns)
+      rdd.values.mapPartitions { iter =>
+        val converter: Array[String] => Row = Conversions.createRowConverter(prunedSchema)
```
JoshRosen: Here, I took the opportunity to do a bit of minor performance optimization by re-using the row converter rather than creating a new one for each row. There are many other inefficiencies / opportunities for optimization in the converter itself, but I'll address those in a separate patch.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -49,50 +49,74 @@ private[redshift] case class RedshiftRelation(
...
-      val unescaped = params.query.map(q => s"($q)").orElse(params.table).get
-      unescaped.replace("'", "\\'")
-    }
     val query = s"SELECT $columnList FROM $tableNameOrSubquery $whereClause"
     val fixedUrl = Utils.fixS3Url(params.tempPath)
 
     s"UNLOAD ('$query') TO '$fixedUrl' WITH CREDENTIALS '$credsString' ESCAPE ALLOWOVERWRITE"
   }
 
-  private def makeRdd(schema: StructType): RDD[Row] = {
```
JoshRosen: Here, I just inlined the bodies of these functions after some refactoring.  I think the new code is slightly easier to follow.

src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
```diff
@@ -240,6 +240,10 @@ class RedshiftIntegrationSuite
       TestUtils.expectedData)
   }
 
+  test("count() on DataFrame created from a Redshift table") {
+    assert(sqlContext.sql("select * from test_table").count() === TestUtils.expectedData.length)
```
marmbrus: This is minor, but even for things like this I'd consider using `checkAnswer` and explicit `COUNT(*)`, just cause when things do go wrong the query plans can be insightful.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -49,50 +49,74 @@ private[redshift] case class RedshiftRelation(
...
     new RedshiftWriter(jdbcWrapper).saveToRedshift(sqlContext, data, updatedParams)
   }
 
-  private def unloadToTemp(columnList: String = "*", whereClause: String = ""): Unit = {
-    val conn = jdbcWrapper.getConnector(params.jdbcDriver, params.jdbcUrl)
-    val unloadSql = unloadStmnt(columnList, whereClause)
-    val statement = conn.prepareStatement(unloadSql)
+  private val tableNameOrSubquery: String = {
+    val unescaped = params.query.map(q => s"($q)").orElse(params.table).get
+    unescaped.replace("'", "\\'")
```
marmbrus: Why is this escaping needed?

marmbrus:
LGTM

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
JoshRosen: This is a carryover from old code, but I should have documented it there. In a nutshell, the issue is that the subquery / table name that is passed to the UNLOAD command ends up being enclosed in single quotes, so any single quotes within that string need to be escaped another time to account for the additional layer of outer single quotes.  I'll add an inline comment to explain this, since it's not obvious at first glance.

src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
JoshRosen: Good idea; will update.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
JoshRosen: Actually, on closer inspection I think that this is actually a mistake when we're handling an empty column query, since in those cases we won't end up adding the extra layer of single quotes. I'm going to add a test to expose this bug and will refactor to fix this.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -49,50 +49,74 @@ private[redshift] case class RedshiftRelation(
...
+      // rather than unloading data.
+      val whereClause = FilterPushdown.buildWhereClause(schema, filters)
+      val countQuery = s"SELECT count(*) FROM $tableNameOrSubquery $whereClause"
+      logInfo(countQuery)
+      val conn = jdbcWrapper.getConnector(params.jdbcDriver, params.jdbcUrl)
+      try {
+        val results = conn.prepareStatement(countQuery).executeQuery()
+        if (results.next()) {
+          val numRows = results.getLong(1)
+          val parallelism = sqlContext.getConf("spark.sql.shuffle.partitions", "200").toInt
```
JoshRosen: Whoops, this is unused. I'll update to pass it into `parallelize`.

JoshRosen:
Added a regression test for the subquery bug, added more comments, fixed the bug, and updated to use checkAnswer, so I'm going to merge this. Thanks for the review.

PR #58, opened 2015-08-27
Allow a single published artifact to work with multiple Hadoop versions

JoshRosen:
This commit allows us to publish one `spark-redshift` artifact which is built against a fixed Hadoop version but which works with both Hadoop 1.x and 2.x. In the past, we published separate artifacts for Hadoop 1.x and 2.x in order to work around a binary incompatibility in `TaskAttemptContext` (see #19). This patch works around the incompatibility using reflection, similar to https://github.com/apache/spark/pull/6599.

In order to make this testable, this patch also modifies our SBT build and Travis configuration so that the test Spark and Hadoop dependencies can be configured separately from the compile dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,9 @@ cache:
 # Spark 1.5.0 only supports Java 7+.
 matrix:
   include:
-    - jdk: openjdk6
-      scala: 2.10.4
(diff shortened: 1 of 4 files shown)


marmbrus:
LGTM

JoshRosen:
Merging this now.

PR #43, opened 2015-08-18
Add support for temporary IAM credentials

eduardoramirez:
Adding support for temporary IAM credentials. Attaching the Security Token (provided by AWS) to the credentials in either COPY or UNLOAD will allow these temporary credentials to work.

More information about [temporary credentials](http://docs.aws.amazon.com/redshift/latest/dg/r_copy-temporary-security-credentials.html).

R=@traviscrawford

PR #59, opened 2015-08-27
Re-add support for AWS STS

eduardoramirez:
Re added the support for AWS STS using a session security token. (original #43)

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ need to be configured to allow access from your driver application.
  </tr>
  <tr>
     <td><tt>aws_security_token</tt></td>
-    <td>No, unless using temporary IAM credentials</td>
+    <td>No, unless using AWS Security Token Service</td>
     <td>No default</td>
     <td>AWS security token corresponding to provided access key.</td>
  </tr>
--- a/src/main/scala/com/databricks/spark/redshift/Parameters.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Parameters.scala
@@ -189,7 +189,14 @@ private[redshift] object Parameters extends Logging {
      */
     def credentialsString(configuration: Configuration): String = {
       val ((_, accessKeyId), (_, secretAccessKey)) = credentialsTuple(configuration)
-      s"aws_access_key_id=$accessKeyId;aws_secret_access_key=$secretAccessKey"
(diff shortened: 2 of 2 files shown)


JoshRosen:
LGTM, so I'll merge this now. Thanks for catching this!

JoshRosen:
@eduardoramirez, do you happen to be using the `s3a` filesystem with STS? I found a StackOverflow answer which suggests that it's unsupported by the `s3` and `s3n` protocols: http://stackoverflow.com/a/28160821/590203

JoshRosen:
Also, one other thing that I'm a bit confused about: due to the bug identified in #60, I don't understand how you were able to get data to/from S3 after loading/unloading it from Redshift. In this patch you only pass the STS token into the `credentialsString` that gets included in Redshift queries, but there doesn't seem to be any explicit code for passing the credential into the Hadoop FileSystem calls that read / write to S3. As a result, I'm assuming that `s3a` is somehow picking up the token from the environment or through `SparkContext.hadoopConfiguration`. If that's the case, though, and it turns out to be the only way to get this to work then I think we need to document that in the README. This might also imply that it would be possible to automatically retrieve the token from the `hadoopConfiguration`.

I'm going to write an STS integration suite and will probably figure out all of this myself, but it would be a huge help if you could confirm some of these theories.

eduardoramirez:
@JoshRosen When keys are omitted, the system uses the EC2 instance's IAM Role for authentication. However, when I ran the jobs locally I did use S3a to set the keys but where my personal dev spare.

The patch was in response to the load/unload parts of Redshift since  `spark-redshift` uses UNLOAD and COPY which require keys as part of the query. Since IAM roles cannot be used when performing such queries to Redshift you need to retrieve STS keys using the EC2 instance's role and attach them as credentials (along with the token): 
http://docs.aws.amazon.com/redshift/latest/dg/r_copy-temporary-security-credentials.html

I did a bit of research looking into how I could incorporate STS into the reading/writing of S3 but found the same information as you. I don't think it's possible to attach the token through Hadoop Conf in any of S3 filesystems. It seems that if you need to use STS it would naturally be from an EC2 instance using IAM roles. I agree that the best thing to do would to document the usage more explicitly in the README to clarify this use case.

PR #60, opened 2015-08-28
Encode the S3 credentials into the URL when accessing S3

marmbrus:
Currently, write jobs to Redshift will fail when the credentials are not put into the global Hadoop configuration. This patch changes `spark-redshift` to encode credentials into the S3 URI for both reads and writes.


JoshRosen:
I'd still like to figure out how / why the existing Travis tests didn't expose this bug.

JoshRosen:
I wonder if Hadoop FileSystem caching is somehow responsible for this. Maybe the credentials that we set for the read are being cached and re-used in the write. I'll see about bypassing the cache when creating filesystems to see if that exposes the problem in Travis.

One gotcha: it might be nice to support AWS security tokens here, if we can (/cc @eduardoramirez).

JoshRosen:
This is actually a bit tricker than I originally thought. Currently, we use a Hadoop `Configuration` to set read credentials but use a URI to encode write credentials. I believe that we could use the URI for both. There may be cases where the URI is not capable of encoding certain types of credentials, such as keys that contain slashes, or is not compatible with certain S3 settings (encryption, perhaps?). In these cases, the only option may be to use the Hadoop configuration.

Since we cannot pass a Hadoop Configuration into the spark-avro writer, the only way to set Configuration options for writes seems to be to modify the SparkContext's global `hadoopConfig`. I think that we can instruct users to do this if they need to use those features but should not have spark-redshift itself be mutating that config.

Therefore, here's my proposal:

- If credentials were not obtained from the Hadoop configuration, then encode them in the URI for both reads and writes.
- Never mutate the `hadoopConfiguration` from `spark-redshift`.
- Add integration tests for working with AWS security token service credentials and figure out whether the URI-based approach can support this.

JoshRosen:
Oh, and for the record the FileSystem caching issue did turn out to be why we didn't observe this issue in Travis.

I figured out how to configure IAM such that our Redshift test account user receives its S3 permissions via an IAM role that is assigned to it. In addition to holding this role, I defined an inline IAM policy on the user which allows it to use the `sts:AssumeRole` command to assume the role that grants the S3 permissions (the same role that the user already holds). Thanks to these policy changes, we should now be able to use the test account's IAM access keys to generate new temporary STS access keys and tokens in order to test out the STS features.  I'm going to take a shot at documenting our test policy configurations and adding STS integration tests in a separate PR.

.travis.yml
```diff
@@ -32,9 +32,9 @@ env:
     # AWS_REDSHIFT_USER
     - secure: "LIkY/ZpBXK3vSFsdpBSRXEsgfD2wDF52X8OZOlyBJOiZpS4y1/obj8b3VQABDPyPH95bGX/LOpM0vVM137rYgF0pskgVEzLMyZOPpwYqNGPf/d4BtQhBRc8f7+jmr6D4Hrox4jCl0cCKaeiTazun2+Y9E+zgCUDvQ8y9qGctR2k="
     # TEST_AWS_ACCESS_KEY_ID
-    - secure: "kFmNLBHXPCIMdbDtHpULFzQtQqWJOO5Q3PtHyvfXE9oHaVwmQUfNd2ygoH5qnKuHO66bfPC7GSY5cMERNIFXsdhDEbRYFj6B4OgcQsYmKAG6b3+qcFYq6htJqkeFbOZB8kUN4YOjlkZg1lPXp/RV0Ac1g72mEsm2CvIoCXmDMYg="
+    - secure: "es6ERiruS8vrv6Ur2GTs5aD3ULcfuGZGJWH/AJhjncVdC5+bkVg2mIMqFjh1sDXlQhUtLXNC9cpg9ZMTBSwn+xYdIr2u8iJZZX3kvmLwKOeR+wsWJpt5WbUPM4eIyZC9tHP9VlL42CJ2AxrHyiGSqSOl1jdMeFMadVUwkmiHg7M="
     # TEST_AWS_SECRET_ACCESS_KEY
-    - secure: "Z34uNm0HuDhyMG9jOhf0nmFR3RKxmF4o1NGg8mNdCuuIykelgY/YDi1LXf5fLz2yQXsmSwt308iq2E1n3Dv6ZBmS4a1s1dpZjTTfjXsyMpz0Hub9ia5B3Th9y0GPjgrMhazFoDf/OQnDxarthJEdJjvTeyo935F5kCK4gikzV8A="
+    - secure: "RAWTY02D5qKITWoroJFxmB6BUmDXe4NfVHhvNTd5BA3mW55w9yrpJoDLOUk97m6IzX9wQJ+E6M7G1v+BqUD4ZN7s6Clb6B8VQxNVSLc/+T0NN4QPonSivIWBtjFwjoErNN763CTPoNm95YD/qGg61q6gWZrKv+ppfZ0qubldjLw="
```
JoshRosen: Our old secret access key contained a slash, hence this change.

JoshRosen:
@marmbrus, I've updated this patch + description to use S3 URI credential encoding for reads and writes. This means that credentials passed via the parameters map or loaded via the default provider chain cannot contain backslashes in the secret key. I did manage to retain support for backslash-containing keys as long as they are provided via `SparkContext.hadoopConfig`. I suppose that I should add an integration test for this (I did test it manually, though); I can do this tomorrow.

JoshRosen:
This is actually going to interact really badly with STS credentials, since the auto-generated temporary credentials are likely to contain slashes (this broke my local STS integration test). As near as I can tell, the only way to make this work in Spark 1.4.x and 1.5.x will be to mutate `SparkContext.hadoopConfiguration`, since `HadoopFsRelation` is hardcoded to always source its Hadoop configurations from there. In addition, we're likely to run into problems with FileSystem caching causing new configurations to be ignored.

In a nutshell, I'm becoming increasingly convinced that there's not going to be a clean backwards-compatible way to manage configurations without involving some sort of global state somewhere. The assumption that you have a single configuration per FS seems to be sort of implicitly baked into some Hadoop APIs (such as FileSystem caching) and we seem to inherit some of those assumptions in Spark.

Given this: should we just go all-in on relying on `SparkContext.hadoopConfiguration` and remove the configuration options in this library, or is there some sort of migration path that would let us avoid contributing to the problem in new libraries like `spark-redshift`?

/cc @marmbrus, @aarondav, and @pwendell, since all of you likely have thoughts here.

JoshRosen:
I am going to close this PR for now and will open a new one to explore some other approaches for testing / fixing some of these credential problems.

Issue #61, opened 2015-08-29
Support writing Decimal columns back to Redshift


JoshRosen:
We should support saving decimals back to Redshift. I think that this might block on adding proper Decimal support to `spark-avro`: https://github.com/databricks/spark-avro/issues/80

PR #62, opened 2015-08-29
Support reading unloaded decimals

JoshRosen:
This patch adds support for reading unloaded decimal columns. Previously, we would fail to parse the string into a decimal, leading to ClassCastExceptions.

### DIFF
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -64,10 +64,10 @@ object SparkRedshiftBuild extends Build {
         "org.scalamock" %% "scalamock-scalatest-support" % "3.2" % "test"
       ),
       libraryDependencies ++= Seq(
-        "org.apache.hadoop" % "hadoop-client" % testHadoopVersion.value % "test",
-        "org.apache.spark" %% "spark-core" % testSparkVersion.value % "test" exclude("org.apache.hadoop", "hadoop-client"),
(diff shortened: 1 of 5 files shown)


JoshRosen:
Note that there are still problems when loading decimals _into_ Redshift; they appear to be truncated to longs. This will be fixed separately (see #61), but may be blocked by adding proper decimal support to `spark-avro`.

JoshRosen:
Note that there are two more things that I need to do before this is merge-worthy:

- Fix / extend support for decimals in the schema-parsing code for the `redshiftFile` API.
- Add tests for reading the largest / smallest possible decimal values and tests for varying precision and scale. I may be able to adapt some test cases from Spark's own decimal tests.

JoshRosen:
Oh no! It looks like stopping and starting new HiveContexts is somehow leaking PermGen space or something. I'm going to try enabling forking in tests and will bump up the permgen limit to see if that quickly fixes the problem.

JoshRosen:
Only remaining task here is adding a comment and adding tests + support in `avroFile`.

JoshRosen:
I think that we're going to deprecate `redshiftFile` in 0.5.0 (and remove it in the next release if nobody complains; see #65), so I'm going to skip fixing the decimal support there. Given this, I'm going to merge this now in order to minimize merge conflicts with a bunch of other in-flight patches. I might loop back later to add additional tests and solicit post-hoc review.

PR #64, opened 2015-08-30
Log warning if tempdir S3 bucket lacks an object lifecycle configuration

JoshRosen:
This patch adds automatic checks which warn users if `tempdir` points to a S3 bucket that lacks an object lifecycle configuration. The goal of this check is to make sure that users realize that these files will need to be manually cleaned up.

I also performed a bit of README cleanup and updated the build file to mark the AWS SDK as a required dependency.

### DIFF
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -55,9 +55,10 @@ object SparkRedshiftBuild extends Build {
       resolvers +=
         "Spark 1.5.0 RC2 Staging" at "https://repository.apache.org/content/repositories/orgapachespark-1141",
       libraryDependencies ++= Seq(
-        "com.amazonaws" % "aws-java-sdk-core" % "1.9.40" % "provided",
-        "com.amazonaws" % "aws-java-sdk-s3" % "1.9.40" % "provided",
-        // We require spark-avro, but avro-mapred must be provided to match Hadoop version:
+        "com.amazonaws" % "aws-java-sdk-core" % "1.9.40",
+        "com.amazonaws" % "aws-java-sdk-s3" % "1.9.40",
+        // We require spark-avro, but avro-mapred must be provided to match Hadoop version.
+        // In most cases, avro-mapred will be provided as part of the Spark assembly JAR.
         "com.databricks" %% "spark-avro" % "1.0.0",
         "org.apache.avro" % "avro-mapred" % "1.7.6" % "provided" exclude("org.mortbay.jetty", "servlet-api"),
         // A Redshift-compatible JDBC driver must be present on the classpath for spark-redshift to work.
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -52,6 +52,7 @@ object SparkRedshiftBuild extends Build {
         "Spark 1.5.0 RC2 Staging" at "https://repository.apache.org/content/repositories/orgapachespark-1141",
       libraryDependencies ++= Seq(
         "com.amazonaws" % "aws-java-sdk-core" % "1.9.40" % "provided",
+        "com.amazonaws" % "aws-java-sdk-s3" % "1.9.40" % "provided",
         // We require spark-avro, but avro-mapred must be provided to match Hadoop version:
         "com.databricks" %% "spark-avro" % "1.0.0",
         "org.apache.avro" % "avro-mapred" % "1.7.6" % "provided" exclude("org.mortbay.jetty", "servlet-api"),
--- a/src/main/scala/com/databricks/spark/redshift/Parameters.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Parameters.scala
@@ -18,7 +18,12 @@ package com.databricks.spark.redshift
 
 import java.net.URI
 
-import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
+import com.amazonaws.services.s3.model.BucketLifecycleConfiguration
+
+import scala.collection.JavaConverters._
+
+import com.amazonaws.auth.{BasicAWSCredentials, DefaultAWSCredentialsProviderChain}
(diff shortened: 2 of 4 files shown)


JoshRosen:
@marmbrus, this is slightly tricky to unit / integration test, so I propose that we postpone 100% perfect test coverage for this feature until we have more time. I do feel that we should get this patch in for 0.5.0 because I think that it will be tricker to change the default configuration validation behavior once we've shipped a release.

project/SparkRedshiftBuild.scala
```diff
@@ -52,6 +52,7 @@ object SparkRedshiftBuild extends Build {
         "Spark 1.5.0 RC2 Staging" at "https://repository.apache.org/content/repositories/orgapachespark-1141",
       libraryDependencies ++= Seq(
         "com.amazonaws" % "aws-java-sdk-core" % "1.9.40" % "provided",
+        "com.amazonaws" % "aws-java-sdk-s3" % "1.9.40" % "provided",
```
JoshRosen: One risk: this adds a new dependency. This reminds me: where do the existing AWS SDK dependencies come from? Spark itself does not have a direct dependency on the AWS SDK and as far as I know neither does Hadoop.

Is this only marked as provided in order to avoid dependency conflicts in certain less-common scenarios? If so, maybe we should consider whether declaring this as a regular dependency would be be easier for more users. We just need to figure out whether dep. conflicts or missing deps are going to be more common here. /cc @jaley, any comments?

src/main/scala/com/databricks/spark/redshift/Parameters.scala
```diff
@@ -242,5 +256,26 @@ private[redshift] object Parameters extends Logging {
...
+    def validateS3Configuration(configuration: Configuration): Unit = {
+      if (!disableS3LifecycleCheck) {
+        val ((_, accessKeyId), (_, secretAccessKey)) = credentialsTuple(configuration)
+        val s3Client = new AmazonS3Client(new BasicAWSCredentials(accessKeyId, secretAccessKey))
+        val s3URI = new AmazonS3URI(Utils.fixS3Url(tempDir))
+        val bucket = s3URI.getBucket
+        val bucketLifecycleConfiguration = s3Client.getBucketLifecycleConfiguration(bucket)
+        val key = s3URI.getKey
+        val someRuleMatchesTempDir = bucketLifecycleConfiguration.getRules.asScala.exists { rule =>
+          rule.getStatus == BucketLifecycleConfiguration.ENABLED && key.startsWith(rule.getPrefix)
```
JoshRosen: Minor subtlety: this only checks that there's an active matching rule, not that the rule actually enforces deletion. Still, this is better than nothing.

project/SparkRedshiftBuild.scala
JoshRosen: Discussed offline. These actually seem necessary and may be unlikely to be provided by the environment, so I'm going to switch these to required dependencies.

JoshRosen:
One concern that I have here: imagine that a new user, eager to try out this library, hits an initial roadblock because they're re-using a temporary directory inside of a companywide S3 bucket. If they're not paying attention and see our error message with a link to the lifecycle management, then I worry that someone might hastily follow the instructions and could enable deletion for a bucket which contains data that they want to keep. This could be an unintended side effect of us _forcing_ users to use best-practices before they can make any progress. For this reason, I'm going to make this into a warning rather than an error.

JoshRosen:
Alright, I've tested this warning manually and it seems to work as expected, so I'm going to merge this now. We can continue iterating later.

project/SparkRedshiftBuild.scala
jaley: I originally marked the AWS SDK as provided for two reasons, though I wasn't 100% sure it's the right thing to do at the time, so we may want to revise. Here's what was bothering me about making them required though:

* On Amazon EMR, which now officially supports Spark pre-installed, the AWS SDK is already on the classpath. EMRFS is also on the classpath, so everything should "just work" with no extra dependencies.
* If you're using S3 and you're not on EMR, you're likely going to be needing `hadoop-aws` instead for the `S3NativeFileSystem` implementation, which has a dependency on the AWS SDK.
* These dependencies are not especially lightweight, and I often find they're a reliable source of dependency hell. I seem to recall them both pulling in Joda, Jackson and Guava, with version ranges/wildcards and all kinds of fun stuff that will no-doubt fight with your Hadoop setup.

JoshRosen: To try to reason through these points:

- I don't think that marking these libraries as required dependencies should be an issue on EMR unless Spark's `userClasspathFirst` setting is enabled, at least in the sense that `spark-redshift`'s dependencies on these libraries should not cause the Amazon-provided versions to be evicted / overridden on the classpath.
- Taking Databricks Cloud as an example of a non-EMR system which interacts with S3, I believe that we also provide `hadoop-aws` and the AWS SDK.

I guess the real concern here is not that we necessarily override the environment's copy of the AWS SDK, but that the SDK's dependency on Joda, Jackson, and Guava ends up pulling in newer versions of those libraries which evict earlier versions that some of the rest of the user-provided libraries might rely on. However, I also wonder if those same users might run into trouble anyways if the environment-provided version of the library causes its versions of Joda, Jackson, and Guava to take precedence when `userClasspathFirst=false`.

In a nutshell, I think this is a decision that could easily go either way. Based on discussion above, I don't think that it's necessarily going to make a difference for Databricks or EMR users; I think the impact would be most heavily concentrated on users with version-conflict-prone dependencies on the SDK's dependencies who deploy with `userClasspathFirst` or deploy in environments where this dependency is not provided.

JoshRosen: EDIT: false -> true

JoshRosen: We've discussed this a little bit more offline and have decided to switch back to marking these dependencies as `provided`, but with an additional comment that captures some of this discussion and a re-wording of the relevant line in the README. I'll submit a PR for this today.

PR #65, opened 2015-09-01
Deprecate several APIs before 0.5.0 release

JoshRosen:
This commit deprecates several APIs in advance of the 0.5.0 release:

- Deprecate the Scala helper functions, since they don't save much typing. Unlike some of the other data sources where the typical configuration consists only of a filename (such as Avro or Parquet), Redshift has several required configuration parameters. These convenience methods are really only saving users from the burden of writing "com.databricks.spark.redshift", which doesn't seem like a huge win.
- Deprecate the variant of `redshiftFile` which accepts a schema string. This code path is not tested and is likely to diverge over time. Most users will be better off using the data sources API or performing the type casts themselves.
- Remove the documentation for the `overwrite` configuration option, which is only used by the deprecated APIs. For consistency / minimizing the testing surface, I think that the code paths for dealing with overwriting should be written with reference to data sources' `SaveMode` API. I'm going to add some additional unit tests for this in a separate followup PR to make sure that the `SaveMode` flags are handled properly.
- Remove the documentation for the other deprecated methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,26 +119,11 @@ OPTIONS (dbtable 'my_table',
          url 'jdbc:redshift://host:port/db?user=username&password=pass');
 ```
 
-### Scala helper functions
-
-The <tt>com.databricks.spark.redshift</tt> package has some shortcuts if you're working directly
-from a Scala application and don't want to use the Data Sources API:
-
-```scala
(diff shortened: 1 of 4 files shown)


JoshRosen:
I'm going to merge this now.

PR #67, opened 2015-09-02
Refactor of AWS credentials handling

JoshRosen:
This patch makes several changes to how `spark-redshift` handles S3 credentials.

The motivation for this patch is a bug where we discovered that credentials were not being set properly for writes. This bug did not occur in our integration tests due to HDFS FileSystem caching behavior: an earlier test ended up creating a FileSystem with valid credentials and that cahced FileSystem was re-used when creating FileSystems in subsequent tests.

Unfortunately, `spark-avro` does not expose a mechanism for passing S3 credentials when writing files, so our only options for passing credentials for writes are to either encode them into the `s3n://` URI or to mutate `SparkContext`'s default `hadoopConfiguration`. Neither of these options are great:
- Due to Hadoop bugs, the URI-based approach will not work for AWS secret keys that contain backslashes.
- Mutating `hadoopConfiguration` is a bad idea because it's a confusing modification of global shared state. If the end user wants to modify this configuration, that's fine, but I feel strongly that libraries should not implicitly modify this configuration. Mutating `hadoopConfiguration` will also prevent users from using multiple sets of AWS keys in the same job, since different modifciations will overwrite each other.

It's non-trivial to extend `spark-avro` with a Hadoop Configuration-passing mechanism because `spark-avro` uses Spark's `HadoopFsRelation`, which hardcodes a dependency on `SparkContext.hadoopConf`.

Given these limitations, this patch modifies `spark-redshift` to use the following approach to handling S3 configurations:
- Remove the configuration parameters for setting AWS access keys, secret keys, and security tokens.
- Rely on the configurations to be provided via sources that are external to this library and its configuration. The Hadoop S3 FileSystems will receive their credentials via system properties, environment variables, URIs, or explicit configuration. If credentials are encoded into a S3 URI then those credentials take precedence over credentials read others mechanism.
- In order to determine which credentials to pass to Redshift, `spark-redshift` will load the credentials from the URI, Hadoop Configuration, or EC2 instance role according to the same rules / precedence used by Hadoop.

There are two corner-cases that are worth mentioning:
- If EC2 IAM authentication is being used, then `spark-redshift` still requires a set of temporary keys and a session token to pass to Redshift. In this case, we still require the user to pass these credentials to us via a set of `sts_` configs ("sts" stands for "security token service", which is an AWS feature).
- If someone is exclusively using the library through the SQL interface (such as over a JDBC / Thriftserver connection to Spark), then the only way to specify S3 credentials is by encoding them in the URI. This precludes the use of secret keys that contain backslashes. Note that the old approach of having configuration parameters for these keys did not solve this problem because we still ended up having to pass the user-provided keys into Hadoop and the only way to do so without modifying the global `hadoopConf` was to re-encode them into the URI (see discussion at #60 for more details).

Finally, there's one key point that I'd like to emphasize: if future versions of Spark and `spark-avro` have a more flexible strategy for handling Hadoop configurations, then it will always be possible to re-introduce _new_ configuration options for setting AWS keys. However, if we were to ship those options today, while we lack the support to implement consistent semantics for them, then we might be painting ourselves into an API compatibility corner.

**TL;DR**: we should be conservative in what APIs we commit to; let's start by only supporting AWS credential configuration mechanisms that have well-understood semantics, since we can always extend things in an additive manner in future releases.


JoshRosen:
By the way, the mega README update is pending soon. Pushing this out now for early feedback.

.travis.yml
```diff
@@ -32,9 +32,9 @@ env:
     # AWS_REDSHIFT_USER
     - secure: "LIkY/ZpBXK3vSFsdpBSRXEsgfD2wDF52X8OZOlyBJOiZpS4y1/obj8b3VQABDPyPH95bGX/LOpM0vVM137rYgF0pskgVEzLMyZOPpwYqNGPf/d4BtQhBRc8f7+jmr6D4Hrox4jCl0cCKaeiTazun2+Y9E+zgCUDvQ8y9qGctR2k="
     # TEST_AWS_ACCESS_KEY_ID
-    - secure: "kFmNLBHXPCIMdbDtHpULFzQtQqWJOO5Q3PtHyvfXE9oHaVwmQUfNd2ygoH5qnKuHO66bfPC7GSY5cMERNIFXsdhDEbRYFj6B4OgcQsYmKAG6b3+qcFYq6htJqkeFbOZB8kUN4YOjlkZg1lPXp/RV0Ac1g72mEsm2CvIoCXmDMYg="
+    - secure: "es6ERiruS8vrv6Ur2GTs5aD3ULcfuGZGJWH/AJhjncVdC5+bkVg2mIMqFjh1sDXlQhUtLXNC9cpg9ZMTBSwn+xYdIr2u8iJZZX3kvmLwKOeR+wsWJpt5WbUPM4eIyZC9tHP9VlL42CJ2AxrHyiGSqSOl1jdMeFMadVUwkmiHg7M="
```
JoshRosen: Note to self: I should actually roll this change back, since I explicitly want to use credentials which contain backslashes to make sure that it will still work if those credentials are provided as part of the Hadoop configuration.

src/it/scala/com/databricks/spark/redshift/STSIntegrationSuite.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import com.amazonaws.auth.BasicAWSCredentials
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient
+import com.amazonaws.services.securitytoken.model.AssumeRoleRequest
+
+import org.apache.spark.sql.{Row, SaveMode}
+import org.apache.spark.sql.types.{IntegerType, StructField, StructType}
+
+/**
+ * Integration tests for accessing S3 using Amazon Security Token Service (STS) credentials.
```
JoshRosen: Important note here: this class actually _does_ test configuration with a set of STS temporary keys that it requests from a real AWS account (I configured the proper roles and policies to allow for this to be done securely in our dev environment). However, this isn't actually testing the more realistic scenario where the EC2 instance's IAM roles are used to authenticate the Spark <-> S3 connection. As a result, it's possible for this suite's ability to detect bugs to be broken during refactoring, since it's theoretically still possible for the Hadoop configuration's EC2 keys to be passed into Redshift if the test author misspelled the `sts` configuration parameter names.

src/it/scala/com/databricks/spark/redshift/STSIntegrationSuite.scala
```diff
@@ -0,0 +1,84 @@
...
+ */
+class STSIntegrationSuite extends IntegrationSuiteBase {
+
+  private val STS_ROLE_ARN: String = loadConfigFromEnv("STS_ROLE_ARN")
```
JoshRosen: Whoops, forgot to commit this to Travis config. Will do shortly.

src/main/java/com/databricks/spark/redshift/S3Credentials.java
```diff
@@ -0,0 +1,100 @@
...
+
+package com.databricks.spark.redshift;
+
+import java.net.URI;
+
+import org.apache.hadoop.conf.Configuration;
+
+/**
+ * <p>
+ * Extracts AWS credentials from the filesystem URI or configuration.
```
JoshRosen: This is copied from Hadoop (with license headers preserved) mostly for convenience and because I didn't want to depend on an unstable / private Hadoop API.

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -20,10 +20,18 @@ import java.io.File
 import java.net.URI
 import java.util.UUID
 
+import scala.collection.JavaConverters._
+import scala.util.control.NonFatal
+
+import com.amazonaws.auth.BasicAWSCredentials
+import com.amazonaws.services.s3.{AmazonS3URI, AmazonS3Client}
+import com.amazonaws.services.s3.model.BucketLifecycleConfiguration
+import org.apache.spark.Logging
```
JoshRosen: We should really remove our dependence on Spark's `Logging` trait; I'll do this in a followup and will add a Scalastyle check to make sure it's not re-introduced.

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -47,4 +55,38 @@ private[redshift] object Utils {
...
+      val bucket = s3URI.getBucket
+      val bucketLifecycleConfiguration = s3Client.getBucketLifecycleConfiguration(bucket)
+      val key = s3URI.getKey
+      val someRuleMatchesTempDir = bucketLifecycleConfiguration.getRules.asScala.exists { rule =>
+        // Note: this only checks that there is an active rule which matches the temp directory;
+        // it does not actually check that the rule will delete the files. This check is still
+        // better than nothing, though, and we can always improve it later.
+        rule.getStatus == BucketLifecycleConfiguration.ENABLED && key.startsWith(rule.getPrefix)
+      }
+      if (!someRuleMatchesTempDir) {
```
JoshRosen: For reviewers: this block of code is just moved from the old file, not changed.

src/main/scala/com/databricks/spark/redshift/AWSCredentials.scala
```diff
@@ -0,0 +1,81 @@
...
+  def credentialsString: String = {
+    val credentials = s"aws_access_key_id=$accessKey;aws_secret_access_key=$secretAccessKey"
+    credentials + sessionToken.map { token => s";token=$token" }.getOrElse("")
+  }
+
+}
+
+object AWSCredentials {
+
+  def load(tempPath: String, hadoopConfiguration: Configuration): AWSCredentials = {
```
JoshRosen: This is the trickiest part of this patch and thus requires a bit of explanation. The idea here is that the Spark <-> S3 credentials have already been configured via the URI, `hadoopConf`, or environment, so we'll attempt to read those credentials so that we can later pass them on to Redshift.

The code below tries to match the logic that the Hadoop S3 FileSystems use while resolving credentials.

src/main/scala/com/databricks/spark/redshift/AWSCredentials.scala
```diff
@@ -0,0 +1,81 @@
...
+    // A good reference on Hadoop's configuration loading / precedence is
+    // https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md
+    val uri = new URI(tempPath)
+    uri.getScheme match {
+      case "s3" | "s3n" =>
+        val creds = new S3Credentials()
+        creds.initialize(uri, hadoopConfiguration)
```
JoshRosen: This case is pretty easy, since we just use the appropriate class from Hadoop.

src/main/scala/com/databricks/spark/redshift/AWSCredentials.scala
```diff
@@ -0,0 +1,81 @@
...
+          val accessKey = hadoopConfiguration.get("fs.s3a.access.key", null)
+          val secretKey = hadoopConfiguration.get("fs.s3a.secret.key", null)
+          if (accessKey != null && secretKey != null) {
+            Some(AWSCredentials(accessKey, secretKey, sessionToken = None))
+          } else {
+            None
+          }
+        }.getOrElse {
+          // Finally, fall back on the instance profile provider
+          val creds = new InstanceProfileCredentialsProvider().getCredentials
```
JoshRosen: One potential bug here, which will require somewhat tricky manual testing: if we're using IAM for the Spark <-> S3 connection and STS for Redshift <-> S3 then we won't be able to get any keys here and an error / exception will probably occur.

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -47,4 +55,38 @@ private[redshift] object Utils {
...
+   * Checks whether the S3 bucket for the given UI has an object lifecycle configuration to
+   * ensure cleanup of temporary files. If no applicable configuration is found, this method logs
+   * a helpful warning for the user.
+   */
+  def checkThatBucketHasObjectLifecycleConfiguration(
+      tempDir: String,
+      awsCredentials: AWSCredentials): Unit = {
+    try {
+      val s3Client = new AmazonS3Client(
+        new BasicAWSCredentials(awsCredentials.accessKey, awsCredentials.secretAccessKey))
```
JoshRosen: Argh, potential bug here: with EC2 IAM credentials we won't have a key here, so `awsCredentials` really needs to be an `Option` here (or we need to change the interface).

project/SparkRedshiftBuild.scala
```diff
@@ -57,6 +57,7 @@ object SparkRedshiftBuild extends Build {
       libraryDependencies ++= Seq(
         "com.amazonaws" % "aws-java-sdk-core" % "1.9.40",
         "com.amazonaws" % "aws-java-sdk-s3" % "1.9.40",
+        "com.amazonaws" % "aws-java-sdk-sts" % "1.9.40" % "test",
```
marmbrus: Why is this one `"test"` but the others are not?

JoshRosen: It's only needed in the integration tests, where we use these APIs to obtain STS credentials for use in the tests.

src/main/scala/com/databricks/spark/redshift/AWSCredentials.scala
marmbrus: This sounds reasonable to me, but please move this to a comment.

project/SparkRedshiftBuild.scala
marmbrus: I see, typically users will just provide the token themselves?

JoshRosen: For now, yes. We might be able to automatically generate the token, but figuring out whether we can/should do that may require PM investigation, I think.

marmbrus: makes sense.

JoshRosen: FWIW, I just spun up an EC2 instance to try a couple of things out and it looks like it's pretty easy to obtain security tokens when the EC2 instance has an IAM profile:

```scala

Comment (JoshRosen) [inline project/SparkRedshiftBuild.scala]: Example code:

```scala
val creds = new InstanceProfileCredentialsProvider().getCredentials.asInstanceOf[AWSSessionCredentials]
STS_ACCESS_KEY_ID = creds.getAWSAccessKeyId
STS_SECRET_ACCESS_KEY = creds.getAWSSecretKey
STS_SESSION_TOKEN = creds.getSessionToken
```

In this case the InstanceProfileCredentialsProvider should wind up creating a background thread to automatically manage refresh of these credentials. So, in a nutshell, it should be _really_ easy for users who use IAM to get this set up: just add this line themselves at the top of their app, then add the configuration to thread these variables into `spark-redshift`'s configuration.

Until we understand all of the complications a little better I don't want to push this into the library itself, but it should be fine to provide this as an example / starter code I think.

JoshRosen: Actually, I don't know that this auto-refreshes with this particular provider... all the more reason to leave this up to the user.

JoshRosen:
Unfortunately, the unit tests are now failing because they ended up using `file://` instead of a real S3 URI in order to test things. I'm going to work around this by splitting out large unit test suite into smaller unit tests which don't run into the problems caused by the end-to-end flavor of the current RedshiftSourceSuite.

JoshRosen:
Alright, this patch should now be ready for review. Please focus on the changes in the README; I'd like feedback to know whether my examples are clear enough for users.

README.md
```diff
@@ -141,9 +141,13 @@ val records = sc.newAPIHadoopFile(
...
 
-You can provide AWS credentials via the parameters listed below, with Hadoop `fs.*` configuration settings, or by making them available via the usual environment variables, system properties or IAM roles.
+There are three ways of configuring AWS credentials for use by this library:
+
+1. Specify AWS keys via [Hadoop configuration properties](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md). For example, if your `tempdir` configuration points to a `s3n://` filesystem then you can set the `fs.s3n.awsAccessKeyId` and `fs.s3n.awsSecretAccessKey` properties in a Hadoop XML configuration file or call `sc.hadoopConfig.set()` to mutate Spark's global Hadoop configuration.
+2. Encode the keys into the `tempdir` URI. For example, the URI `s3n://ACCESSKEY:SECRETKEY@bucket/path/to/temp/dir` encodes the key pair (`ACCESSKEY`, `SECRETKEY`). Due to Hadoop limitations, this approach will not work for secret keys which contain forward slash (`/`) characters.
```
JoshRosen: "Hadoop limitations" probably ought to link to the relevant JIRA.

README.md
```diff
@@ -141,9 +141,13 @@ val records = sc.newAPIHadoopFile(
...
+`spark-redshift` reads and writes data to S3 when transferring data to/from Redshift. As a result, it requires AWS credentials with read and write access to a S3 bucket (specified using the `tempdir` configuration parameter). Assuming that Spark has been configured to access S3, `spark-redshift` should automatically discover the proper credentials to pass to Redshift.
 
-You can provide AWS credentials via the parameters listed below, with Hadoop `fs.*` configuration settings, or by making them available via the usual environment variables, system properties or IAM roles.
+There are three ways of configuring AWS credentials for use by this library:
+
+1. Specify AWS keys via [Hadoop configuration properties](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md). For example, if your `tempdir` configuration points to a `s3n://` filesystem then you can set the `fs.s3n.awsAccessKeyId` and `fs.s3n.awsSecretAccessKey` properties in a Hadoop XML configuration file or call `sc.hadoopConfig.set()` to mutate Spark's global Hadoop configuration.
```
JoshRosen: Out of user empathy, I should probably add some copy-paste-able examples of mutating `sc.hadoopConf` in Scala for both the `s3n` and `s3a` filesystems.

JoshRosen: And in Python as well, since it's a little less obvious how to do things there.

README.md
```diff
@@ -141,9 +141,13 @@ val records = sc.newAPIHadoopFile(
...
+2. Encode the keys into the `tempdir` URI. For example, the URI `s3n://ACCESSKEY:SECRETKEY@bucket/path/to/temp/dir` encodes the key pair (`ACCESSKEY`, `SECRETKEY`). Due to Hadoop limitations, this approach will not work for secret keys which contain forward slash (`/`) characters.
+3. If you are running on EC2 and authenticate to S3 using IAM and [instance profiles](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html), then you must must configure the `temporary_aws_access_key_id`, `temporary_aws_access_key_id`, and `temporary_aws_session_token` configuration properties to point to temporary keys created via the AWS [Security Token Service](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html). These temporary keys will then be passed to Redshift via `LOAD` and `UNLOAD` commands.
```
JoshRosen: I think that I actually have some example code which might be able to auto-generate these credentials. I'm tempted to just dump it in here both as a reference to myself and to get more eyes on it so that any bugs are fixed.

JoshRosen:
Actually, I'm going to merge this now and will defer the additional README improvements to be incorporated in a more general README auditing / editing pass before the 0.5.0 release. I'll incorporate any post-hoc review comments in followup PRs.

README.md
kyprifog: Should there be a line added here to the readme about avoiding logging the aws credentials in the UNLOAD ... WITH CREDENTIALS statement of the aws credentials when authenticating by this method

PR #71, opened 2015-09-04
Remove dependency on Spark's Logging trait

### LINKED ISSUE #69: Remove dependency on spark.Logging
As a best-practice, third-party code should not rely on Spark's `Logging` trait. We should remove this dependency and should add a Scalastyle rule to make sure that it's not re-introduced in new code.

JoshRosen:
As a best-practice, third-party code should not rely on Spark's `Logging` trait. This patch refactors the code to remove this dependency and adds a Scalastyle rule to make sure that it's not re-introduced in new code.

Fixes #69.

### DIFF
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -55,6 +55,8 @@ object SparkRedshiftBuild extends Build {
       resolvers +=
         "Spark 1.5.0 RC3 Staging" at "https://repository.apache.org/content/repositories/orgapachespark-1143",
       libraryDependencies ++= Seq(
+        // The SLF4J API is provided by Spark:
+        "org.slf4j" % "slf4j-api" % "1.7.5" % "provided",
         // These Amazon SDK depdencies are marked as 'provided' in order to reduce the risk of
         // dependency conflicts with other user libraries. In many environments, such as EMR and
         // Databricks, the Amazon SDK will already be on the classpath. In other cases, the SDK is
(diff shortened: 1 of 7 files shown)


scalastyle-config.xml
```diff
@@ -123,6 +123,11 @@ This file is divided into 3 sections:
       // scalastyle:on println]]></customMessage>
   </check>
 
+  <check customId="sparklogging" level="error" class="org.scalastyle.file.RegexChecker" enabled="true">
+    <parameters><parameter name="regex">org.apache.spark.Logging</parameter></parameters>
+    <customMessage><![CDATA[Do not rely on Spark's Logging trait.]]></customMessage>
```
marmbrus: Nice.

project/SparkRedshiftBuild.scala
```diff
@@ -55,6 +55,8 @@ object SparkRedshiftBuild extends Build {
       resolvers +=
         "Spark 1.5.0 RC3 Staging" at "https://repository.apache.org/content/repositories/orgapachespark-1143",
       libraryDependencies ++= Seq(
+        // The SLF4J API is provided by Spark:
+        "org.slf4j" % "slf4j-api" % "1.7.5" % "provided",
```
marmbrus: This probably doesn't matter since this library will always be included with Spark, but I believe you are actually supposed to include this jar.  That way it gracefully fails even if the user has not included a specific slf4j binding.

marmbrus:
Minor comment, but LGTM.

project/SparkRedshiftBuild.scala
JoshRosen: I noticed that `spark-csv` also marks it as provided, but I think you're right: this should be included because it doesn't have any dependencies of its own. I'll update to do this now, since it would be better to set a good example here.

PR #74, opened 2015-09-05
Use a fresh temporary directory for each scan

JoshRosen:
This patch fixes a bug where multiple scans of the same relation would end up sharing the same temporary directory in S3, causing them to overwrite each others' data. This, in turn, led to exceptions during query execution.

This patch fixes this bug by creating a fresh subdirectory inside of the temporary directory for each scan.


JoshRosen:
@marmbrus, this was the cause of the bugs that we saw while running TPCDS.

JoshRosen:
Come to think of it, the assumption that we're using a separate directory for each scan and unload can probably be exploited elsewhere... I wonder if there's an explicit overwrite flag which we can remove so that we'll get errors if we ever end up accidentally sharing the directory.

JoshRosen:
Going to merge this now.

PR #75, opened 2015-09-08
Update spark-avro dependency

marmbrus:


JoshRosen:
```
java.lang.ClassCastException: org.apache.hadoop.mapreduce.Job cannot be cast to org.apache.hadoop.mapred.JobContext
	at org.apache.hadoop.mapred.OutputCommitter.setupJob(OutputCommitter.java:213)
	at org.apache.spark.sql.sources.BaseWriterContainer.driverSideSetup(commands.scala:351)
	at org.apache.spark.sql.sources.InsertIntoHadoopFsRelation.insert(commands.scala:157)
	at org.apache.spark.sql.sources.InsertIntoHadoopFsRelation.run(commands.scala:139)
	at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult$lzycompute(commands.scala:57)
	at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult(commands.scala:57)
	at org.apache.spark.sql.execution.ExecutedCommand.doExecute(commands.scala:68)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:88)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:88)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:147)
	at org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:87)
	at org.apache.spark.sql.SQLContext$QueryExecution.toRdd$lzycompute(SQLContext.scala:950)
	at org.apache.spark.sql.SQLContext$QueryExecution.toRdd(SQLContext.scala:950)
```

Looks like a symptom of a possible Hadoop version mismatch?

JoshRosen:
I'm thinking that this could actually be a classloading issue because otherwise the error message doesn't make a lot of sense to me. The error is `org.apache.hadoop.mapreduce.Job cannot be cast to org.apache.hadoop.mapred.JobContext`, but `Job` implements `mapreduce.JobContext`, which in turn extends `mapred.JobContext`.

JoshRosen:
Spotted the problem and fixed things, so I'm going to merge this fix now. We can bump to 2.0.1 in a separate commit.

PR #78, opened 2015-09-09
Refactor handling of SaveModes and overwrite parameter

### LINKED ISSUE #68: Audit SaveMode codepaths and internals
See comments that I left in #65; we need to audit the places where the `overwrite` configuration is used and need to make sure that this matches up with the semantics of the `SaveMode.overwrite` (and possibly unify code paths).

JoshRosen:
This patch performs refactoring and cleanup related to `spark-redshift`'s handling of SaveModes and its `overwrite` configuration parameter.

- The `overwrite` parameter is no longer used internally; instead, it's used at the edges of the API to pick an appropriate `SaveMode` setting. This is important because this parameter is deprecated and is slated to be removed in 0.6.0.
- Use a transaction to atomically swap the temporary table and backup table when performing overwrites with a staging table.

Fixes #68.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -28,7 +28,7 @@ import scala.util.control.NonFatal
 import com.databricks.spark.redshift.Parameters.MergedParameters
 
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+import org.apache.spark.sql.{DataFrame, Row, SaveMode, SQLContext}
 import org.apache.spark.sql.types._
 
 /**
@@ -116,11 +116,12 @@ private[redshift] class RedshiftWriter(
   private def doRedshiftLoad(
       conn: Connection,
       data: DataFrame,
+      saveMode: SaveMode,
       params: MergedParameters,
       tempDir: String): Unit = {
 
     // Overwrites must drop the table, in case there has been a schema update
-    if (params.overwrite) {
+    if (saveMode == SaveMode.Overwrite) {
       val deleteExisting = conn.prepareStatement(s"DROP TABLE IF EXISTS ${params.table.get}")
       deleteExisting.execute()
     }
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -85,28 +85,27 @@ private[redshift] class RedshiftWriter(
...
-      log.info("Loading new Redshift data to: " + tempTable)
-      log.info("Existing data will be backed up in: " + backupTable)
-
       action(tempTable)
+
       if (jdbcWrapper.tableExists(conn, table)) {
-        conn.prepareStatement(s"ALTER TABLE $table RENAME TO $backupTable").execute()
+        conn.prepareStatement(
+          s"""
+             | BEGIN;
```
JoshRosen: @jaley, was there any reason that you chose to not use a transaction to perform these renames atomically?

src/main/scala/com/databricks/spark/redshift/package.scala
```diff
@@ -76,12 +76,19 @@ package object redshift {
...
-     * Load the DataFrame into a Redshift database table
+     * Load the DataFrame into a Redshift database table. By default, this will append to the
+     * specified table. If the `overwrite` parameter is set to `true` then this will drop the
+     * existing table and re-create it with the contents of this DataFrame.
      */
     @deprecated("Use DataFrame.write()", "0.5.0")
     def saveAsRedshiftTable(parameters: Map[String, String]): Unit = {
       val params = Parameters.mergeParameters(parameters)
-      DefaultRedshiftWriter.saveToRedshift(dataFrame.sqlContext, dataFrame, params)
+      val saveMode = if (params.overwrite) {
```
JoshRosen: Here, and reviewers should confirm this, the old non-overwrite behavior was to append, since we'd just end up performing a copy into the existing table.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
jaley: I seem to remember thinking that Redshift mostly ignores transactions. Looking at the docs now though, I think that's only the case for table truncation and it should be fine for these renames. Definitely a better option if it seems to be working?

JoshRosen: Yeah, it passes the integration tests. This seems marginally safer than the approach that was there before because it eliminates the (remote) possibility of a fatal error occurring between the separate steps when renaming the tables.

I think that this approach is marginally cleaner, too, since it lets me change the scope of the try-finally block.

jaley: Sounds good, I agree with you.

JoshRosen:
Actually, I just noticed one potential problem: this code path is exercised by unit tests, but I don't think that we have any integration tests for the case where you overwrite an existing table. There's a regression test for a bug which occurred when overwriting a non-existent table, but that will go through a different branch than the one which contains the transaction.

Before we commit this, I need to add a new integration test to make sure that we're properly covering that branch.

JoshRosen:
I added an integration test for overwrite with an existing table and it looks like it passed, so I'm going to go ahead and merge these changes.

Issue #84, opened 2015-09-12
Column names cannot contain spaces when saving back to Redshift


JoshRosen:
Because `spark-redshift` uses Avro in its write path, it inherits the limitations of Avro's schema validation: we cannot create tables with columns that contain non-letter or `_` characters, such as spaces or quotes.

In order to work around these limitations, I think that we'd have to use a different set of column names when writing to Avro, then use a [JSONPaths](https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-json-jsonpaths) file to map those columns back to the original column names.

l15k4:
Hey, colons too, I have fields like this `c-geo:c3` and getting : 
```
[info]   org.apache.avro.SchemaParseException: Illegal character in: c-geo:c3
[info]   at org.apache.avro.Schema.validateName(Schema.java:1083)
```

Spec https://avro.apache.org/docs/current/spec.html#names doesn't say anything about this. Wdyt ?

JoshRosen:
The Avro spec actually _does_ say something about this (emphasis mine to make things clearer):

> Record, enums and fixed are named types. Each has a _fullname_ that is composed of two parts; a _name_ and a _namespace_. Equality of names is defined on the fullname.
>
> The name portion of a fullname, record field names, and enum symbols must:
>
> * start with `[A-Za-z_]`
> * **subsequently contain only `[A-Za-z0-9_]`**

My PR description also notes this:

> we cannot create tables with columns that contain non-letter-or-`_` characters, such as spaces or quotes.

Starting in version `3.0.0-preview1`, you can use the `csv` `tempformat` during writes and that format has none of these restrictions on names. In that version, there's even a helpful error message which detects when you've hit this case using AVRO serialization and suggests the CSV tempformat as a workaround: https://github.com/databricks/spark-redshift/blob/6cc49da71d7abd62dd3bb4491c34b7010fa886f5/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala#L370

I'll mark this issue as "documentation" to remind me to add a section to the README documenting this workaround.

JamieL22:
Is this fixed yet? (white space in column names)

PR #85, opened 2015-09-12
Quote column names in JDBC schemaString

### LINKED ISSUE #80: Reserved words cannot be used as column names when writing back to Redshift
`spark-redshift` does not currently support the use of reserved words as column names when creating tables in Redshift. For example, `table` cannot be used as a column name.

This should be simple to fix: simply quote all column names when constructing the create table statement in `RedshiftJDBCWrapper.schemaString`.

JoshRosen:
This patch modifies `JDBCWrapper.schemaString` to wrap column names in quotes, which is necessary in order to allow us to create tables with columns whose names are reserved words or which contain spaces. This fixes #80.

Note that, by itself, this patch does not enable full support for creating Redshift tables with column names that contain spaces; we are currently constrained by Avro's schema validation rules (see #84).

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -336,6 +336,14 @@ class RedshiftIntegrationSuite extends IntegrationSuiteBase {
         .mode(SaveMode.ErrorIfExists)
         .save()
       assert(DefaultJDBCWrapper.tableExists(conn, tableName))
+      val loadedDf = sqlContext.read
+        .format("com.databricks.spark.redshift")
+        .option("url", jdbcUrl)
+        .option("dbtable", tableName)
+        .option("tempdir", tempDir)
+        .load()
+      assert(loadedDf.schema === df.schema)
+      checkAnswer(loadedDf, Seq(Row(1)))
     } finally {
       conn.prepareStatement(s"drop table if exists $tableName").executeUpdate()
       conn.commit()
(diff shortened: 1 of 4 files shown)


JoshRosen:
@marmbrus, do you think that we should perform similar quoting in Spark SQL's built-in JDBC datasource? Is this type of quoting a dialect-specific thing? These questions aren't blockers to making this change here in `spark-redshift`, but I just wanted to briefly consider those questions to make sure that we're not overlooking potential bugs in Spark.

marmbrus:
Yeah, this is a known issue in Spark ([SPARK-9505](https://issues.apache.org/jira/browse/SPARK-9505)) as well.  However there I think we will have to work it into dialects as I think different systems use different quoting mechanisms (at least Spark SQL is different than MySQL).

Do we need to do the same thing when querying such columns?  or do we already escape there?

rxin:
The JDBC dialect dev API already has a quoting mechanism defined.



On Sep 13, 2015, at 12:00 PM, Michael Armbrust <notifications@github.com>
wrote:

Yeah, this is a known issue in Spark (SPARK-9505
<https://issues.apache.org/jira/browse/SPARK-9505>) as well. However there
I think we will have to work it into dialects as I think different systems
use different quoting mechanisms (at least Spark SQL is different than
MySQL).

Do we need to do the same thing when querying such columns? or do we
already escape there?

—
Reply to this email directly or view it on GitHub
<https://github.com/databricks/spark-redshift/pull/85#issuecomment-139908329>
.

JoshRosen:
@marmbrus, we already wrap in quotes when querying; it looks like we were just missing support for this when creating tables.

JoshRosen:
Going to merge this now.

JoshRosen:
Added an unload to this test, just to make it clear that the read path is also covered.

PR #88, opened 2015-09-14
Refactor RedshiftSourceSuite to remove use of ScalaMock

JoshRosen:
This patch refactors RedshiftSourceSuite to remove the use of ScalaMock.

The motivation for this is the fact that our previous ScalaMock code became a bit of an impediment when trying to refactor codepaths which created and closed new JDBC connections.


JoshRosen:
Discussed offline; will merge now to unblock another PR.

PR #90, opened 2015-09-15
Automatically use Redshift 4.0 or 4.1 JDBC driver, depending on which is installed

### LINKED ISSUE #83: Automatically use Redshift 4.0 or 4.1 JDBC driver, depending on which is installed
Currently, the default value of `jdbcdriver` is `com.amazon.redshift.jdbc4.Driver`, but this will not work if a user has installed the JDBC 4.1 driver. Instead of hardcoding either the 4.0 or 4.1 driver, we should handle the absence of a user-provided `jdbcdriver` config by first trying to load the JDBC4 driver and falling back to attempt the 4.1 driver (or vice versa).

See https://github.com/databricks/spark-redshift/issues/56#issuecomment-139690983 for additional context.

JoshRosen:
This patch refactors the driver-loading code so that it automatically uses either the Redshift JDBC 4.0 or 4.1 drivers, depending on which is installed. I also added support for automatically supplying an appropriate default JDBC driver class name when the `postgres://` subprotocol is used.

I tested all of the error-handling corner-cases manually.

Fixes #83.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,8 +239,8 @@ and use that as a temp location for this data.
  <tr>
     <td><tt>jdbcdriver</tt></td>
     <td>No</td>
-    <td><tt>com.amazon.redshift.jdbc4.Driver</tt></td>
-    <td>The class name of the JDBC driver to load before JDBC operations. Must be on classpath.</td>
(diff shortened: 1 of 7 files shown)


marmbrus:
LGTM

PR #93, opened 2015-09-23
Fix Utils.joinURLs on Windows

JoshRosen:
This PR fixes an issue where `spark-redshift` would not work properly on Windows because the `Utils.joinUrls` method implicitly used the platform-specific path separator via `File`.


JoshRosen:
Going to merge this now.

PR #94, opened 2015-09-23
Remove binary dependency on org.apache.hadoop.fs.s3.S3FileSystem

### LINKED ISSUE #92: `hadoop-aws` dependency
We're running `spark-redshift` on EMR 4. We're trying to avoid including `org.apache.hadoop:hadoop-aws` in our build, as it conflicts with many things already provided by the EMR environment.

The assertion [here](https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/Utils.scala#L112), ironically, seems to require that the `S3Filesystem` class is available on the classpath just so that we can assert that it isn't being used? We're seeing the following stacktrace on the `0.5.0` release:

```
java.lang.NoClassDefFoundError: org/apache/hadoop/fs/s3/S3FileSystem
	at com.databricks.spark.redshift.Utils$.assertThatFileSystemIsNotS3BlockFileSystem(Utils.scala:104)
	at com.databricks.spark.redshift.RedshiftWriter.saveToRedshift(RedshiftWriter.scala:273)
	at com.databricks.spark.redshift.DefaultSource.createRelation(DefaultSource.scala:104)
	at org.apache.spark.sql.sources.ResolvedDataSource$.apply(ddl.scala:309)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:144)
...
```

On EMR, that class shouldn't be available, as we don't want to use it. We instead use the `EmrFileSystem` class, which is provided by the environment already.

JoshRosen:
This removes an unintended binary dependency on the `org.apache.hadoop.fs.s3.S3FileSystem` class. Our use of this class introduced an implicit dependency on the `hadoop-aws` artifact, which is not present in EMR.

Fixes #92.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Utils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Utils.scala
@@ -27,7 +27,6 @@ import com.amazonaws.services.s3.{AmazonS3URI, AmazonS3Client}
 import com.amazonaws.services.s3.model.BucketLifecycleConfiguration
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.FileSystem
-import org.apache.hadoop.fs.s3.S3FileSystem
 import org.slf4j.LoggerFactory
 
 /**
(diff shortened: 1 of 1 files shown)


jaley:
Thanks for taking care of that so fast!

Cc: @emlyn as he ran into his earlier today.

PR #98, opened 2015-09-29
Fix problems in mapping between Spark and Redshift types for short and float columns

JoshRosen:
This patch fixes two data-type-related problems which occur in roundtrip saves and loads of float- and short-typed columns:
- When creating Redshift tables, Spark SQL's `ShortType` was mapped to Redshift's `INTEGER` type instead of `SMALLINT`. When reading Redshift tables back into DataFrames, `SMALLINT` was mapped to `IntegerType` instead of `ShortType`.
- When creating DataFrames from Redshift tables, `getCatalystType()` mapped Redshift's `REAL` into `DoubleType`, but it actually corresponds to `FloatType`.

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -287,7 +287,10 @@ class RedshiftIntegrationSuite extends IntegrationSuiteBase {
         .option("dbtable", tableName)
         .option("tempdir", tempDir)
         .load()
-      assert(loadedDf.schema === TestUtils.testSchema)
+      // Redshift doesn't have a BYTE type, so these schemas won't match for that column:
+      assert(
+        loadedDf.schema.filterNot(_.name == "testbyte") ===
+        TestUtils.testSchema.filterNot(_.name == "testbyte"))
       checkAnswer(loadedDf, TestUtils.expectedData)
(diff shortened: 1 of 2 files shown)


JoshRosen:
While I think that this is the right fix to make here, this is technically breaking a type of compatibility. Therefore, I'd like to put this patch on hold until we decide to cut a 0.6.0 release.

JoshRosen:
If we're going to do 3.0.0 then we might as well fix this so that we can simplify some of the roundtrip tests. OTOH I suppose that this risks breakage of ETL pipelines which convert Redshift to Parquet since the resulting Parquet schema might change.

JoshRosen:
Given the high risk of breakage to ETL pipelines, I'm going to mark this as won't fix for now. If users feel especially strongly that this must be fixed, then I'll add a fix and will guard it behind a feature flag. If you want that behavior, then please explicitly comment here (not just a reaction on my comment) so that I know.

Nasreen23:
> Given the high risk of breakage to ETL pipelines, I'm going to mark this as won't fix for now. If users feel especially strongly that this must be fixed, then I'll add a fix and will guard it behind a feature flag. If you want that behavior, then please explicitly comment here (not just a reaction on my comment) so that I know.

I think this fix is strongly recommended. The spark data frame is bringing back Integer type for a SMALL INT sql type, which is causing an issue while trying to persist to redshift via parquet.

PR #99, opened 2015-09-29
Use manifest to skip COPY of empty partitions when writing to Redshift

### LINKED ISSUE #96: Avro schema only file causes Redshift stl_load_errors: "Invalid AVRO file found. Unexpected end of AVRO file."
When a part file contains no data and just the Avro schema, Redshift throws an exception:

```
Caused by: com.amazon.support.exceptions.ErrorException: [Amazon](500310) 
Invalid operation: Load into table 'stg_device_attribute' failed.  Check 
'stl_load_errors' system table for details.;
```

Where the load error reads:

```
Invalid AVRO file found. Unexpected end of AVRO file.
```

My AVRO file just contains the schema for a number of shards where no records matched a `WHERE`-clause condition:

```
Obj...avro.codecsnappy.avro.schema..{"type":"record","name":"topLevelRecord","fields":[{"name":"roster_id","type":["long","null"]},{"name":"device_id","type":["long","null"]},{"name":"push_token","type":["string","null"]},{"name":"shard_id","type":["long","null"]},{"name":"device_type","type":["string","null"]},{"name":"device_date_created","type":["string","null"]}]} 
```

In my particular case I am running a query across 12 PostgreSQL databases which may not contain results. I can get around this problem by merging into a smaller number of partitions and hope there's at least one record in each part file by sacrificing Redshift parallelism on the load.

I am not sure what the solution is but it might be worth it to warn if a part file will have 0 records.

JoshRosen:
This WIP patch fixes #96, an issue where Redshift's COPY command did not cope well with empty Avro partitions.

The fix implemented here is to use a [manifest](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) to instruct Redshift to load only the non-empty partitions' Avro files.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -141,7 +141,8 @@ private[redshift] class RedshiftWriter(
       // See https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html
       // for a full description of the manifest format.
       val manifestEntries = filesToLoad.map { filename =>
-        s"""{"url":"${Utils.fixS3Url(filename)}", "mandatory":true}"""
+        val url = Utils.fixS3Url(Utils.removeCredentialsFromURI(URI.create(filename)).toString)
+        s"""{"url":"$url", "mandatory":true}"""
       }
       val manifest = s"""{"entries": [${manifestEntries.mkString(",\n")}]}"""
       val fs = FileSystem.get(URI.create(tempDir), data.sqlContext.sparkContext.hadoopConfiguration)
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -21,8 +21,11 @@ import java.sql.{Connection, Date, SQLException, Timestamp}
 
 import com.amazonaws.auth.AWSCredentials
 import com.amazonaws.services.s3.AmazonS3Client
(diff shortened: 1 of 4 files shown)


JoshRosen:
Alright, this should now be ready for review.  /cc @marmbrus and @cfeduke for comments.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -132,51 +159,53 @@ private[redshift] class RedshiftWriter(
...
-                 |Table name: ${params.table.get}
-                 |Column name: ${results.getString("colname").trim}
-                 |Column type: ${results.getString("type").trim}$columnLength
-                 |Raw line: ${results.getString("raw_line")}
-                 |Raw field value: ${results.getString("raw_field_value")}
-                """.stripMargin
-            Some(new SQLException(exceptionMessage, e))
-          } else {
-            None
+    manifestUrl.foreach { manifestUrl =>
```
JoshRosen: The only change in this block was to wrap the actual `COPY` statement itself in this `foreach` block in order to skip the copy if no non-empty Avro partitions were written.

cfeduke:
I will run this patch with the same queries that were causing issues tonight; specifically I know of a query that actually causes 0 rows to be returned (something from last Thanksgiving) which would be a decent full coverage test.

JoshRosen:
@cfeduke, did you end up getting a chance to try this out? I'm fairly confident that this should fix the issue, given the regression tests that I added, but more evidence from a real use-case never hurts.

marmbrus:
LGTM

JoshRosen:
Alright, going to merge this now (given that it has integration tests).

PR #101, opened 2015-09-30
Tutorial

sameeraxiomine:
Created the following set of files

1. tutorial/README.md
2. SparkRedshiftTutorial.scala
3. images/loadreadstep.gif
4. images/loadunloadstep.gif
5. images/savetoredshift.gif


tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+ * 1. AWS Access Key
+ * 2. AWS Secret Access Key
+ * 3. Redshift Database Name
+ * 4. Redshift UserId
+ * 5. Redshift Password
+ * 6. Redshift URL (Ex. swredshift.czac2vcs84ci.us-east-1.redshift.amazonaws.com:5439
+ * 
+ */
+object SparkRedshiftTutorial {
+  /*
```
JoshRosen: The spacing and indentation seems off in this file. Is it indented using a mixture of spaces and tabs? Please re-indent using only spaces.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+   * 3. After step 2 %HADOOP_HOME/bin should contain winutils.exe
+   * 
+   */   
+	if (System.getProperty("os.name").startsWith("Windows")) {
+		val hdphome: File = new File("C:/MySoftware/hdphome");
+		System.setProperty("hadoop.home.dir", hdphome.getAbsolutePath().toString())
+	}
+
+	def main(args: Array[String]): Unit = {
+		val awsAccessKey = args(0)
```
JoshRosen: Maybe add a check for the length of `args` and print a simple usage string if it doesn't math the expected length?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+		val rsDbName = args(2)
+		val rsUser = args(3)
+		val rsPassword = args(4)
+    /** Sample Redshift URL is swredshift.czac2vcs84ci.us-east-1.redshift.amazonaws.com:5439 */
+		val rsURL = args(5)  
+		val jdbcURL = s"""jdbc:redshift://$rsURL/$rsDbName?user=$rsUser&password=$rsPassword"""
+    println(jdbcURL)
+		val sc = new SparkContext(new SparkConf().setAppName("SparkSQL").setMaster("local"))
+
+		val tempS3Dir = "s3n://redshift-spark/temp/"
```
JoshRosen: This would have to be configurable if anyone else was going to run this, right?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+		sc.hadoopConfiguration.set("fs.s3n.awsAccessKeyId", awsAccessKey)
+		sc.hadoopConfiguration.set("fs.s3n.awsSecretAccessKey", awsSecretKey)
+
+		val sqlContext = new org.apache.spark.sql.SQLContext(sc)
```
JoshRosen: Why not just import SQLContext at the top of the file so that you don't have to write the fully-qualified name here?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+			.format("com.databricks.spark.redshift")
+			.option("url", jdbcURL) 
+			.option("tempdir", tempS3Dir) 
+			.option("dbtable", "event") 
+			.load()
+		eventsDF.show()
+		eventsDF.printSchema()
+
+    /** Load from a query */
+		val salesQuery = """select salesid,listid,sellerid,buyerid,
```
JoshRosen: Mind capitalizing the SQL keywords, just for consistency, and putting spaces between the column names?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+
+		val eventQuery = "select * from event"
+		val eventDF = sqlContext.read
+			.format("com.databricks.spark.redshift")
+			.option("url", jdbcURL) 
+			.option("tempdir", tempS3Dir)
+			.option("query", eventQuery)
+			.load()
+
+    /** Register 'event' table as 'myevent' in the Spark Environment */
```
JoshRosen: Using the two-star `/**` syntax for inline comments seems odd, since this isn't Java/scaladoc. Mind using `//`-style comments for those instead?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+     */
+		sqlContext.sql("select * from myevent where eventid>1000").withColumnRenamed("eventid", "id")
+			.write.format("com.databricks.spark.redshift")
+			.option("url", jdbcURL) 
+			.option("tempdir", tempS3Dir)
+			.option("dbtable", "redshiftevent")
+			.mode(SaveMode.Append) 
+			.save()
+
+		/** Demonstration of Inter-operability */
```
JoshRosen: "interoperability" is usually written as one word.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+		val salesAGGQuery = """select sales.eventid as id,sum(qtysold) as totalqty,sum(pricepaid)  as salesamt
```
JoshRosen: Spacing is inconsistent here (two spaces before the `as` instead of one). These details matter for public-facing example code.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+		val salesAGGDF2 = salesAGGDF.join(eventsDF, salesAGGDF("id") === eventsDF("eventid"))
+			.select("id", "eventname", "totalqty", "salesamt")
+
+    /** Create a custom schema */
+		val schema = StructType(List(StructField("eventid", IntegerType, false),
+			StructField("eventname", StringType, false),
+			StructField("qtysold", LongType, false),
+			StructField("revenue", DecimalType(38, 2), false)))
+      
+    /** Apply custom schema on existing DataFrame */
```
JoshRosen: This custom schema is essentially just changing the nullability and names of fields, but not the types, right? I wonder if this is going to be confusing to readers because it's not demonstrating a `spark-redshift` feature itself, per-se.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+   * For Windows Users only
+   * 1. Download contents from link 
+   *      https://github.com/srccodes/hadoop-common-2.2.0-bin/archive/master.zip
+   * 2. Unzip the file in step 1 into your %HADOOP_HOME%/bin. 
+   *        In the example below  %HADOOP_HOME%=C:/MySoftware/hdphome
+   * 3. After step 2 %HADOOP_HOME/bin should contain winutils.exe
+   * 
+   */   
+	if (System.getProperty("os.name").startsWith("Windows")) {
+		val hdphome: File = new File("C:/MySoftware/hdphome");
```
JoshRosen: Should this be hardcoded or should be it be loaded from the `HADOOP_HOME` environment variable?

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Source code accompanying the spark-redshift tutorial. 
+ * The following parameters need to be passed
+ * 1. AWS Access Key
+ * 2. AWS Secret Access Key
+ * 3. Redshift Database Name
+ * 4. Redshift UserId
+ * 5. Redshift Password
+ * 6. Redshift URL (Ex. swredshift.czac2vcs84ci.us-east-1.redshift.amazonaws.com:5439
```
JoshRosen: There's an opening left paren on this line but no corresponding closing paren.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.databricks.spark.redshift.tutorial
+
+import org.apache.spark.{ SparkConf, SparkContext }
+import org.apache.spark.sql.SaveMode
+import java.io.File
+import org.apache.spark.sql.types.{ StructType, StructField,
```
JoshRosen: Please take a look at our code style guide to see how we format / organize imports in Spark, then update this to match.

tutorial/SparkRedshiftTutorial.scala
JoshRosen: Also, what is "Spark Environment" here? It seems more straightforward to explain that this is registering a temporary table so that it can be queried via the `sqlContext.sql` interface, etc.

tutorial/SparkRedshiftTutorial.scala
```diff
@@ -0,0 +1,172 @@
...
+			.format("com.databricks.spark.redshift")
+			.option("url", jdbcURL) 
+			.option("tempdir", tempS3Dir)
+			.option("query", eventQuery)
+			.load()
+
+    /** Register 'event' table as 'myevent' in the Spark Environment */
+		eventsDF.registerTempTable("myevent") 
+
+    /** Save to a redshift table from a table registered in spark */
```
JoshRosen: I think that Spark should be capitalized wherever it appears.

JoshRosen: Similarly, Redshift is a product name and should also be capitalized.

JoshRosen:
Not to nitpick, but I think PNG might be a better file format for images due to its better quality / compression ratios.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+When you start the Redshift service you will first need to create the TICKT database and load it. Follow the instructions [here](http://docs.aws.amazon.com/redshift/latest/dg/cm-dev-t-load-sample-data.html) to create/load the TICKT database.
+
+I used a Redshift database running on a 2 node cluster. Each node manages 2 slices for a total of 4 slices. This *usually* (read Redshift [documentation](http://docs.aws.amazon.com/redshift/latest/dg/c_high_level_system_architecture.html) for a more nuanced discussion) means that each table will be stored in 4 separate partitions, one for each slice. 
+
+## Usage##
+
+We are ready to interact with Redshift using the spark-redshift library. The skeleton of the program we will be using is shown in Listing 1. The entire `SparkRedshiftTutorial.scala` program can be accessed from [here](SparkRedshiftTutorial.scala). You can also use the Spark REPL to run the lines listed in the program below. 
+
+```scala
+package com.databricks.spark.redshift.tutorial
```
JoshRosen: If the full tutorial source is available as a `.scala` file, do you think we can cut down on some of the skeleton / harness here to make the prose read a bit more smoothly?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+    val sqlContext = new org.apache.spark.sql.SQLContext(sc)
+  }
+} 
+```
+Listing 1 : SparkRedshiftExamples.scala skeleton
+
+We need the following set of user provided parameters to communicate with AWS in general and Redshift in particular -
+
+- **AWS Access Key and AWS Secret Access Key** - This key pair will be used to communicate with AWS services. This information is passed by the AWS Client libraries in every interaction with AWS.
+- **Redshift Database Name** - When you provision the Redshift service you have to provide a name of for your database. This is similar to a schema in Oracle. The name of my Redshift database was sparkredshift
```
JoshRosen: I'd probably wrap sparkredshift in markdown backticks, similar to the O'Reilly etc. style for drawing attention to string literals in their prose.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+               .option("tempdir", tempS3Dir) 
+               .option("dbtable", "event")
+               .load()
+eventsDF.show()   
+```
+Listing 2 : Reading an entire Redshift table 
+
+`.format("com.databricks.spark.redshift")` line tells the DataSource API that we are using the `spark-redshift` package. It uses this information to identify the class `DefaultSource` in the package specified by the `format` invocation. This class contains the entry points for the functionality provided by the Data Sources API implementation.
+
+Next we provide the parameters necessary to read the `event` table. We provide the JDBC URL, the temporary S3 folder which will be used to copy the table data to and the table name we want to read. A comprehensive list of parameters is listed on the spark-redshift documentation [page](https://github.com/databricks/spark-redshift#parameters).
```
JoshRosen: Minor nit, but I'd put all of `"spark-redshift documentation page" in the clickable portion of the link. Friendlier to screen-readers and more obviously clickable.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+	val tempS3Dir = "s3n://redshift-spark/temp/"
+    sc.hadoopConfiguration.set("fs.s3n.awsAccessKeyId", awsAccessKeyId)
+    sc.hadoopConfiguration.set("fs.s3n.awsSecretAccessKey", awsSecretAccessKey)
+
+	/*Create the SQL Context*/    
+    val sqlContext = new org.apache.spark.sql.SQLContext(sc)
+  }
+} 
+```
+Listing 1 : SparkRedshiftExamples.scala skeleton
```
JoshRosen: I'm not sure whether `Listing XXX` style makes sense for a Markdown-formatted tutorial document like this, since it's kind of tricky to get the listing labels to appear clearly. Can we restructure this to avoid the need for this style of listing?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+We can even register the DataFrame instance as a temporary table in Spark and execute queries against it as follows
+
+```scala
+eventsDF.registerTempTable("myevent")
+val myEventDF = sqlContext.sql("select * myevent")
+```
+
+While the above examples used Scala we could have also used SQL. An example of writing to a Redshift table in SQL is show below
```
JoshRosen: "is show" -> "is shown"

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+```{r, engine='bash'}
+> %sql
```
JoshRosen: `%sql` only makes sense in DBC, right?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+Figure 1 shows the the actions required to perform this step.
+
+![](images/loadunloadstep.gif)
+
+Figure 1 : UNLOAD action
+
+First the Spark Driver communicates with the Redshift Leader node to obtain the schema of the table (or query) requested. The attribute `override lazy val schema: StructType` in the class `com.databricks.spark.redshift.RedshiftRelation` will obtain the schema on demand by invoking the method `resolveTable` of the class `com.databricks.spark.redshift.JDBCWrapper`. The `JDBCWrapper` class is responsible for fetching the schema from the Redshift Leader.   
+
+Next a Redshift [UNLOAD](http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) query is created using the schema information obtained. The UNLOAD command unloads each slice into a s3 folder (22c365b4-13cb-40fd-b4d6-d0ac5d426551) created in the temporary s3 location (s3n://spark-redshift/temp/) provided by the user. Each file contains a row per line and each column of the row is pipe (|) delimited. This process occurs in parallel for each slice. The spark-redshift library achieves its high performance through this mechanism.
```
JoshRosen: S3 should be capitalized.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+#### UNLOAD Redshift to S3 ####
+
+Figure 1 shows the the actions required to perform this step.
+
+![](images/loadunloadstep.gif)
+
+Figure 1 : UNLOAD action
+
+First the Spark Driver communicates with the Redshift Leader node to obtain the schema of the table (or query) requested. The attribute `override lazy val schema: StructType` in the class `com.databricks.spark.redshift.RedshiftRelation` will obtain the schema on demand by invoking the method `resolveTable` of the class `com.databricks.spark.redshift.JDBCWrapper`. The `JDBCWrapper` class is responsible for fetching the schema from the Redshift Leader.
```
JoshRosen: First, comma.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+### Under the hood - Load Function ###
+
+In this section we will take a peek inside spark-redshift to understand how exactly the LOAD function works. Specifically we will look at how `event` table was read into a DataFrame. The LOAD is a two step process
+
+1. [UNLOAD](http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) data from the Redshift table to S3
+2. Consume the unloaded files in S3 via a custom InputFormat into an RDD which is then wrapped as a DataFrame instance using the schema obtained from Redshift.
+
+#### UNLOAD Redshift to S3 ####
+
+Figure 1 shows the the actions required to perform this step.
```
JoshRosen: Read out of context, this sentence doesn't make sense, since "this step" is ambiguous.

I'd prefer to be more explicit here and say something like "Figure 1" or "The following diagram shows the steps that are performed when Spark Redshift UNLOADs data from Redshift to S3." That sentence makes sense even if the reader didn't look at the section header.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+#### Read UNLOAD'ed S3 files into a DataFrame instance####
+
+Figure 2 shows how the files unloaded in S3 are consumed to form a DataFrame instance
+
+![](images/loadreadstep.gif)
+Figure 2 : Read from S3 into a DataFrame  
+
+Once the files are written to S3, a custom InputFormat implemented in the class `com.databricks.spark.redshift.RedshiftInputFormat` is used to consume the files in parallel. This class is similar to the standard and well known `TextInputFormat` where the key is the byte offset of the start of each line in the file. The value class however, is of type `Array[String]` (unlike `TextInputFormat`where it is `Text`). The value instance is created by splitting the lines using the default delimiter |. The `RedshiftInputFormat` consumes the S3 files line by line to produce an `RDD`. The schema obtained earlier is applied on this `RDD` to generate an instance of a `DataFrame`.
```
JoshRosen: This first sentence would be an awesome opportunity to use more commas, since "implemented in the class `blahblahblah`" is kind of a parenthetical.

Similarly, " using the default delimiter |" needs either backticks around the pipe character, a comma before it, or both.

In the last sentence, you say "an RDD" or "this RDD", but then use the slightly-awkward "an instance of a DataFrame." Why not just say "a DataFrame"?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+          .mode(SaveMode.Append)
+          .save()
+```
+Listing 5: Writing to Redshift tables
+
+spark-redshift automatically creates a Redshift table with the appropriate schema determined from the table/DataFrame being written. The default behavior is to create a new table and to throw an error message if a table with the same name already exists (corresponding to `SaveMode.ErrorIfExists`)
+
+There are two key points to note in Listing 5-
+
+1. Note the `.withColumnRenamed("eventid", "id")`. This feature if necessary if any of the source table columns names are key words in Redshift (ex. table) then spark-redshift will throw and error. This will be [fixed](https://github.com/databricks/spark-redshift/issues/80) in the next release
```
JoshRosen: Note that we might very well release 0.5.1 by the time that we publish the post, in which case we'll want to update this.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+spark-redshift automatically creates a Redshift table with the appropriate schema determined from the table/DataFrame being written. The default behavior is to create a new table and to throw an error message if a table with the same name already exists (corresponding to `SaveMode.ErrorIfExists`)
+
+There are two key points to note in Listing 5-
+
+1. Note the `.withColumnRenamed("eventid", "id")`. This feature if necessary if any of the source table columns names are key words in Redshift (ex. table) then spark-redshift will throw and error. This will be [fixed](https://github.com/databricks/spark-redshift/issues/80) in the next release 
+2. Note how we utilize the modes. The first write utilizes the mode `SaveMode.Overwrite` which means that the table will be dropped and recreated it exists. The second query utilizes `SaveMode.Append` which will create the table if it does not exist but will append to it if it does. The default mode is `SaveMode.ErrorIfExists` which creates the table if it does not exists and throws and error if it does. The last mode is `SaveMode.Ignore` which is same as `SaveMode.Overwrite` if the table does not exist but does nothing if the table exists.
```
JoshRosen: I hate the word "utilize" and wish it would be banned from all technical writing. It tends to give writing an awkward formality and is almost never better than "use" in my opinion.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+While the above examples used Scala we could have also used SQL. And example of writing to a Redshift table in SQL is shows in Listing 6. The only thing to be aware of in the SQL mode is that SaveMode's are not available and only the default mode (`SaveMode.ErrorIfExists`) is in force.
```
JoshRosen: > "And example of writing to a Redshift table in SQL is shows in Listing 6."

Typos.

JoshRosen: "is in force" is a weird phrase; I'd find something simpler.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+USING com.databricks.spark.redshift
+OPTIONS (
+dbtable 'redshiftevent',
+tempdir 's3n://redshift-spark/temp/',
+  url 'jdbc:redshift://swredshift.czac2vcs84ci.us-east-1.redshift.amazonaws.com:5439/sparkredshift?user=spark&password=mysecretpass'
+)
+AS SELECT * FROM myevent;
+```
+
+By default the save operation uses the key distribution [style](http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html) of `EVEN` in Redshift. This can be changed by the use of option parameter `diststyle` and `distkey`. See spark-redshift [documentation](https://github.com/databricks/spark-redshift) for details.
```
JoshRosen: Put all of "key distribution style" in the link.

JoshRosen: "By the use of" -> "using". I'm a Strunk & White fan.

JoshRosen: I'd put all of `spark-redshift` documentation in the link.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+![](images/savetoredshift.gif)
+
+The sequence of steps involved in performing the save function are-
+
+1. The data from the `DataFrame` instance is written out in parallel to the temporary S3 folder. spark-avro is used to write the data to ensure that it encapsulates the schema.
+2. Next a Redshift [COPY](http://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command is created and invoked to load the files created in the temporary S3 folder are loaded it into a Redshift table which respects the value of the `SaveMode` parameter passed.
+
+
+## Integration multiple Datasources ##
```
JoshRosen: Integration _of_ multiple Datasources?

Inte_grating_ multiple Datasources?

Should Datasources be one word or two? How should it be capitalized? Let's be consistent.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+Data read via spark-redshift is automatically converted to `DataFrame` objects, Spark’s primary abstraction for large datasets. This promotes interoperability between data sources, since types are automatically converted to Spark’s standard representations(for example `StringType`, `DecimalType`). A Redshift user can, for instance, join Redshift tables with data stored in S3, Hive tables, CSV or Parquet files stored on HDFS. This flexibility is important to users with complex data pipelines involving multiple sources.
+
+To illustrate this concepts consider Listing 6
```
JoshRosen: Typo. _these_

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+The sequence of steps involved in performing the save function are-
+
+1. The data from the `DataFrame` instance is written out in parallel to the temporary S3 folder. spark-avro is used to write the data to ensure that it encapsulates the schema.
+2. Next a Redshift [COPY](http://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command is created and invoked to load the files created in the temporary S3 folder are loaded it into a Redshift table which respects the value of the `SaveMode` parameter passed.
+
+
+## Integration multiple Datasources ##
+
+Data read via spark-redshift is automatically converted to `DataFrame` objects, Spark’s primary abstraction for large datasets. This promotes interoperability between data sources, since types are automatically converted to Spark’s standard representations(for example `StringType`, `DecimalType`). A Redshift user can, for instance, join Redshift tables with data stored in S3, Hive tables, CSV or Parquet files stored on HDFS. This flexibility is important to users with complex data pipelines involving multiple sources.
```
JoshRosen: "DataFrame objects" -> how about just "DataFrames"?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+             .option("tempdir", tempS3Dir) 
+             .option("dbtable", "redshift_sales_agg") 
+             .mode(SaveMode.Overwrite)
+             .save()
+```
+Listing 7: Providing Custom Schema
+
+Note how we have specified every column as non-nullable in the custom schema.
+
+In this section we illustrated how Data Sources API provides a unified API to work with multiple datasources. This feature is enabled by the generic `DataFrame` interface. We also looked at how a `DataFrame` can be saved to Redshift with a user-defined schema. None of this is unique to spark-redshift. It is provided by the DataSources API for use with all compatible packages. This is the feature that makes the DataSource API such a powerful tool for ETL usecases.
```
JoshRosen: "This is the feature that makes" -> "These features make"

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Under the hood - Putting it all together##
+
+As we discussed earlier Spark SQL will introspect for a class called `DefaultSource` in the Datasources API package `com.databricks.spark.redshift`. The `DefaultSource` class implements the trait `RelationProvider` class which is provides the default load functionality for the library. It consumes the parameters provided by the user and converts it to an instance of `BaseRelation` which is implemented by the class `com.databricks.spark.redshift.RedshiftRelation`.
```
JoshRosen: > the trait `RelationProvider` class

Is it a trait or a class?

> It consumes the parameters provided by the user and converts it to an instance of `BaseRelation` which is implemented by the class `com.databricks.spark.redshift.RedshiftRelation`. 

What is the second "it" referring to?

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Under the hood - Putting it all together##
+
+As we discussed earlier Spark SQL will introspect for a class called `DefaultSource` in the Datasources API package `com.databricks.spark.redshift`. The `DefaultSource` class implements the trait `RelationProvider` class which is provides the default load functionality for the library. It consumes the parameters provided by the user and converts it to an instance of `BaseRelation` which is implemented by the class `com.databricks.spark.redshift.RedshiftRelation`. 
+
+This `com.databricks.spark.redshift.RedshiftRelation` class is responsible for providing an `RDD` of `org.apache.spark.sql.Row` which backs the `org.apache.spark.sql.DataFrame` instance. This represents the underlying implementation for the Load functionality for the spark-redshift package where the schema is inferred from the underlying Redshift table. The functionality for performing load where the user provides the schema is supported by the trait `org.apache.spark.sql.sources.SchemaRelationProvider` and is implemented in the class `RedshiftRelation` .
```
JoshRosen: Why is "Load" capitalized in this first sentence?

> "The functionality for performing load where the user provides the schema" 

This is a kind of wordy phrase and is also passive-voice. I'd like to re-work this sentence to avoid this.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+The store functionality of the spark-redshift package is supported by the trait `org.apache.spark.sql.sources.CreatableRelationProvider` implemented by the `RedshiftWriter`.
+
+
+## Conclusion ###
+
+Data Sources API provides a unified interface for handling structured data in Spark. It was introduced in Spark 1.2 through the spark-sql package. It unifies structured data sources under a common interface which is richer than the original RDD based approach which utilized InputFormat API of Hadoop. In this article we explored how the spark-redshift package expands the Data Sources API to support Amazon Redshift which is becoming increasingly popular choice in the Enterprise IT infrastructure.
```
JoshRosen: _The_ data sources API? Spark's data sources API? With the latter, I'd drop the "in Spark" at the end of this sentence.

> "through the spark-sql package"

This seems unnecessary and potentially-confusing.

"RDD based" should be hyphenated.

> "utilized InputFormat API of Hadoop"

Again, no "utilized". How about "Hadoop's InputFormat API"?

JoshRosen: > becoming increasingly popular choice in the Enterprise IT infrastructure.

Missing a word.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+               .option("query", salesAGGQuery)
+               .load()   
+
+   
+val salesAGGDF2=salesAGGDF.join(eventsDF, salesAGGDF("id") === eventsDF("eventid"))
+                         .select("id", "eventname","totalqty","salesamt")
+```
+Listing 6: Working with DataFrame instances
+
+While both `eventDF` and `salesAGGDF` table are sourced from Redshift, assume for the sake of illustration that they are sourced from a non-Redshift datasource. This should not be too far-fetched since the DataFrame class is part of the Data Sources API and not the spark-redshift library. A DataFrame can be created from any data source compatible with the Data Sources API
```
JoshRosen: This final sentence is missing a period.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+By default the save operation uses the key distribution [style](http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html) of `EVEN` in Redshift. This can be changed by the use of option parameter `diststyle` and `distkey`. See spark-redshift [documentation](https://github.com/databricks/spark-redshift) for details.
+
+### Under the hood - Save Function ###
+
+The implementation of the Save function is provided in the class `com.databricks.spark.redshift.RedshiftWriter`.   
+
+![](images/savetoredshift.gif)
+
+The sequence of steps involved in performing the save function are-
```
JoshRosen: This is a kind of roundabout sentence. How about "The save function performs the following steps:"

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
+#Tutorial#
+
+The [Spark Data Sources API](https://databricks.com/blog/2015/01/09/spark-sql-data-sources-api-unified-data-access-for-the-spark-platform.html) introduced in Spark 1.2 supports a pluggable mechanism for integration with structured data-sources. It is a unified API designed to support two major operations:
+
+1. Loading structured data from an external data source into Spark
+2. Storing structured data from Spark into an external data source.
+
+The Data Sources API has built in integration for several data sources such as Hive, Avro, Json, JDBC and Parquet. Third party integration is added through spark-packages. One such integration is for Amazon Redshift which is provided by the spark-redshift package.
```
JoshRosen: JSON is usually in all-caps, in my experience, since it's an acronym. Third-party should be hyphenated.

tutorial/README.md
```diff
@@ -0,0 +1,413 @@
...
+2. Storing structured data from Spark into an external data source.
+
+The Data Sources API has built in integration for several data sources such as Hive, Avro, Json, JDBC and Parquet. Third party integration is added through spark-packages. One such integration is for Amazon Redshift which is provided by the spark-redshift package. 
+
+Prior to the introduction of spark-redshift, Spark’s JDBC data source was the only way for Spark users to read data from Redshift. While this method is adequate when running queries returning a small number of rows (order of 100’s), it is too slow when handling large scale data. The reason being that JDBC provides a ResultSet based approach where rows are retrieved in a single thread in small batches. Furthermore, the use of JDBC to store large datasets in Redshift is only practical when data needs to be moved between tables inside a Redshift database. The JDBC based INSERT/UPDATE queries are only practical for small updates to Redshift tables. For users hoping to load or store large volumes of data from Redshift, JDBC leaves much to be desired in terms of performance and throughput.
```
JoshRosen: "The reason being that" -> I don't like this phrase.

JoshRosen:
I took one editing pass, but may have additional feedback. I would try reading the current draft aloud to find odd phrasing, typos, and misspellings, then take an editing pass to fix the mechanical issues.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: I have fixed this.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed as below
    if (args.length < 6) {
      println("Needs 6 parameters only passed " + args.length)
      println("parameters needed - $awsAccessKey $awsSecretKey $rsDbName $rsUser $rsPassword $rsURL")
    }

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: I removed this while improving on the comment. My bad. It is enough to pass -D-Dhadoop.home.dir=C:/MySoftware/hdphome in my case. I removed it. Looks cleaner now

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Done

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Done. Also fixed in the blog entry on google docs

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed as per suggested.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Sometimes it is not available when testing in the local environment. This is typical in a Windows environment. The only file needed is the winutils.exe and the correct setting of the hadoop.home.dir (Ex. -D-Dhadoop.home.dir=C:/MySoftware/hdphome). I can skip it but I have (and other I know have too) spent an inordinate amount of time figuring this out. Especially the fact that simply giving a relative path is not enough and an absolute path is needed. Hence added the comments around it.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Fixed both

tutorial/README.md
sameeraxiomine: Agreed. Cut down the comments as they are included in the .scala program.

tutorial/README.md
sameeraxiomine: Done!

tutorial/README.md
sameeraxiomine: Done

tutorial/README.md
sameeraxiomine: Agreed. It was originally the blog so I carried it forward. Fixed it.

tutorial/README.md
sameeraxiomine: Fixed

tutorial/README.md
sameeraxiomine: Yes removed it. Had copied as is from the PDF you had sent me.

tutorial/README.md
sameeraxiomine: Fixed

tutorial/README.md
sameeraxiomine: Agreed. I also got rid of "Figure xxx" notation as the comment for Listing xxx would also apply here.

tutorial/README.md
sameeraxiomine: Very nuanced. Thank you. I fixed it.

tutorial/README.md
sameeraxiomine: I qualified that it is only applicable for versions upto 0.5.0.

tutorial/README.md
sameeraxiomine: :+1:  Fixed! :-)

tutorial/README.md
sameeraxiomine: "Strunk & White fan" - Didn't know what that was. Spent an inordinate time online trying to make sense of it :-). Fixed it as you suggested :-).

tutorial/README.md
sameeraxiomine: Fixed the header as - Integration with other Data Sources

For the rest if I am referring to the API I will call it Data Sources API. If I am simple referring to a source of data I will call it "datasource". If I want to refer to multiple ones I will call it "data sources". Does that sound good.

tutorial/README.md
sameeraxiomine: Agree. I generally struggle with these when I have to parenthesize them. I wrote as follows - `DataFrame`'s. Looks a little funky to me.

tutorial/README.md
sameeraxiomine: Don't see it. Can you clarify

tutorial/README.md
sameeraxiomine: Fixed

tutorial/README.md
sameeraxiomine: Checked scaladoct, it is a trait. I reworded as follows
As we discussed earlier Spark SQL will introspect for a class called `DefaultSource` in the Datasources API package, `com.databricks.spark.redshift`. The `DefaultSource` class implements the trait, `RelationProvider` which provides the default load functionality for the library. The interface provided by the `RelationProvider` trait consumes the parameters provided by the user and converts it to an instance of `BaseRelation` which is implemented by the class, `com.databricks.spark.redshift.RedshiftRelation`.

tutorial/README.md
JoshRosen: "Concepts" is plural; "this" is singular.

tutorial/README.md
sameeraxiomine: Reworded as follows - The load function which supports the a user-defined schema is supported by the trait `org.apache.spark.sql.sources.SchemaRelationProvider` and implemented in the class `RedshiftRelation` .

tutorial/README.md
sameeraxiomine: Fixed

tutorial/README.md
sameeraxiomine: Yes. Fixed

tutorial/README.md
sameeraxiomine: Changed it to - 
This is because, JDBC provides a ResultSet based approach where rows are retrieved in a single thread in small batches.

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Agreed. It does not add any value from the point of view of spark-redshift. Especially since Redshift does not care about the not-null directives and we can change the column name in the SELECT with an AS clause on columns. I re-wrote it as follows

While both `eventDF` and `salesAGGDF` table are sourced from Redshift, assume for the sake of illustration that they are sourced from a non-Redshift datasource. This should not be too far-fetched since the DataFrame class is part of the Data Sources API and not the `spark-redshift` library. A `DataFrame` can be created from any datasource compatible with the Data Sources API.

`salesAGGDF2` `DataFrame` is created by joining `eventsDF` and `salesAGGDF2`. We register it as a temporary table `redshift_sales_agg` before saving it to Redshift with the same name  `redshift_sales_agg`
  

```scala
salesAGGDF2.registerTempTable("redshift_sales_agg")

sqlContext.sql("SELECT * FROM redshift_sales_agg")
          .write.format("com.databricks.spark.redshift")
          .option("url", jdbcURL)
          .option("tempdir", tempS3Dir)
          .option("dbtable", "redshift_sales_agg")
          .mode(SaveMode.Overwrite)
          .save()
```

tutorial/SparkRedshiftTutorial.scala
sameeraxiomine: Can you show where. I see closing paren everwhere

tutorial/README.md
sameeraxiomine: Also changed as gif to png

tutorial/README.md
sameeraxiomine: Re-worded as follows 
We could have achieve similar results using SQL. The only thing to be aware of in the SQL mode is that all the SaveMode's are not available and only the default mode (`SaveMode.ErrorIfExists`) is applicable.

```sql

tutorial/README.md
sameeraxiomine: Thanks - Updated as follows
The following code snippet illustrates these concepts-

tutorial/SparkRedshiftTutorial.scala
JoshRosen: On this line, there's an open paren in "(Ex. " but no closing one.

sameeraxiomine:
I updated the tutorial and source code based on your comments. I made some more of my own as I did a full pass through it.

JoshRosen:
I have some additional comments that I'd like to address, but I'm going to take care of them myself by submitting a followup PR. Therefore, I'm going to merge this now. Thanks!

PR #102, opened 2015-09-30
Fix bug in overwrite when using schema-qualified table names

### LINKED ISSUE #97: [Amazon](500310) Invalid operation: syntax error at or near ".";
When using `SaveMode.Overwrite` (which I honestly haven't checked to see if its supported) like the following code:

```
.toDF().write
      .option("jdbcdriver", "com.amazon.redshift.jdbc41.Driver")
      .option("url", redshiftJdbcUrl)
      .option("dbtable", "pf.some_table_here")
      .option("tempdir", tempDirBucket)
      .mode(SaveMode.Overwrite)
      .save()
```

I encounter:

```
Caused by: com.amazon.support.exceptions.ErrorException: [Amazon](500310) Invalid operation: syntax error at or near ".";
        ... 80 more
```

(Lost the rest of the stack trace in a closed screen session but I believe the error can be reproduced by specifying the schema name with the table name and using `SaveMode.Overwrite`.)

Using `SaveMode.Append` is a work around (I just truncate the table prior to saving).

JoshRosen:
This patch fixes #97, an issue where using `SaveMode.Overwrite` with schema-qualified table names (such as `my_schema.my_table`) would lead to "Invalid operation: syntax error at or near ".";" messages from Redshift.

The problem is that the `ALTER TABLE ... RENAME` command only accepts unqualified table names as the new names, so we need to strip off the schema prior to performing the renaming.

In order to keep this code clean, this patch introduces a new `TableName` case class which manages all of the escaping and parsing logic for table names, then uses that class to fix the Overwrite issue.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/TableName.scala
+++ b/src/main/scala/com/databricks/spark/redshift/TableName.scala
@@ -16,6 +16,8 @@
 
 package com.databricks.spark.redshift
 
+import scala.collection.mutable.ArrayBuffer
+
 /**
  * Wrapper class for representing the name of a Redshift table.
  */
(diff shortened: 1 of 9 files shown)


src/main/scala/com/databricks/spark/redshift/TableName.scala
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Parses a table name which is assumed to have been escaped according to Redshift's rules for
+   * delimited identifiers.
+   */
+  def parseFromEscaped(str: String): TableName = {
+    def dropOuterQuotes(s: String) =
+      if (s.startsWith("\"") && s.endsWith("\"")) s.drop(1).dropRight(1) else s
+    def unescapeQuotes(s: String) = s.replace("\"\"", "\"")
+    def unescape(s: String) = unescapeQuotes(dropOuterQuotes(s))
+    str.split('.').toSeq match {
```
JoshRosen: Argh, it turns out that it's legal for `.` to appear inside of delimited identifiers, so I'll have to do a quote-aware splitting here.

JoshRosen:
My experience in working with a wrapper for `TableName` makes me want to use similar wrappers to automatically handle column name quoting in the other parts of `spark-redshift`. I'll leave that refactoring to future work, though.

marmbrus:
LGTM

PR #104, opened 2015-10-06
Update AWS Java SDK to 1.10.22

JoshRosen:
This upgrades the AWS Java SDK from 1.9.40 to 1.10.22. Although this version is only relevant in testing, I'm upgrading because older versions of the SDK have bugs in the `getBucketLifecycleConfiguration` configuration. According to an automated email from AWS:

> Your account has been identified as a user of the AWS SDK for Java making API requests to Amazon S3's getBucketLifecycleConfiguration operation. With the recent launch of the Standard – Infrequent Access storage class, getBucketLifecycleConfiguration's response can now have a new valid value for the StorageClass attribute: "STANDARD_IA".
>
> The AWS SDK for Java models this response element as an Enum, and STANDARD_IA was not a valid Enum value in previous versions of the SDK. Therefore, if a lifecycle policy involving the new storage class is configured on your bucket through other means such as the AWS Management Console, all getBucketLifecycleConfiguration calls made by previous versions of the AWS SDK for Java will fail with an AmazonClientException. If you plan to utilize Amazon S3's lifecycle feature with the new Standard - Infrequent Access storage class, it is critical that you update your AWS SDK for Java to 1.10.19 or later to avoid running into this error.

This upgrade should help other developers who test this library to avoid receiving the same message from Amazon.

### DIFF
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -64,9 +64,10 @@ object SparkRedshiftBuild extends Build {
         // There's a trade-off here and we've chosen to err on the side of minimizing dependency
         // conflicts for a majority of users while adding a minor inconvienece (adding one extra
         // depenendecy by hand) for a smaller set of users.
-        "com.amazonaws" % "aws-java-sdk-core" % "1.10.22" % "provided",
-        "com.amazonaws" % "aws-java-sdk-s3" % "1.10.22" % "provided",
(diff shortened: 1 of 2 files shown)


JoshRosen:
Ugh, looks like a Jackson conflict:

```
com.fasterxml.jackson.databind.JsonMappingException: Could not find creator property with name 'id' (in class org.apache.spark.rdd.RDDOperationScope)
```

PR #105, opened 2015-10-09
Enable data source short name for Spark 1.5.0+

JoshRosen:
This patch allows `spark-redshift` to take advantage of the new support for data source short names / aliases added in Spark 1.5.0+ (https://github.com/apache/spark/pull/7802). This allows users to write

```
sqlContext.read.format("redshift")
```

instead of

```
sqlContext.read.format("com.databricks.spark.redshift")
```

The short name also works in other contexts, including SQL queries.

Unfortunately, this patch is made overly complicated by a binary compatibility issue: if a data source is compiled against Spark 1.5.0 and extends the new `DataSourceRegister` class, then it will hit ClassNotFoundExceptions when running against Spark 1.4.x. In order to work around this, I inlined a copy of the `DataSourceRegister` class into `spark-redshift` itself.

This compatibility issue also affected `spark-csv`: https://github.com/databricks/spark-csv/pull/160

/cc @pwendell, @marmbrus, @falaki, @jdrit

### DIFF
--- a/project/SparkRedshiftBuild.scala
+++ b/project/SparkRedshiftBuild.scala
@@ -43,7 +43,7 @@ object SparkRedshiftBuild extends Build {
       organization := "com.databricks",
       scalaVersion := "2.10.5",
       crossScalaVersions := Seq("2.10.5", "2.11.7"),
-      sparkVersion := "1.4.1",
+      sparkVersion := "1.5.0",
       testSparkVersion := sys.props.get("spark.testVersion").getOrElse(sparkVersion.value),
       testHadoopVersion := sys.props.get("hadoop.testVersion").getOrElse("2.2.0"),
       spName := "databricks/spark-redshift",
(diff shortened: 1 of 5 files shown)


JoshRosen:
When running on Spark 1.4.x, this should be fine because they'll only be one version of the `DataSourceRegister` class available at runtime.

When running on 1.5.x or newer, however, there will be two versions of this class on the classpath: the version from `spark-redshift` and Spark's built-in version. If `spark.driver.userClassPathFirst` or `spark.executor.userClassPathFirst` are set first, however, I could imagine problems arising in case some contexts use `spark-redshift`'s copy of the class and others use Spark's built-in copy. We might also have similar issues via other mechanisms which allow the user classpath to take precedence over Spark's own classes, such as `--driver-class-path`.

Before we can merge this, I'm going to need more time to think through and test this.

JoshRosen:
At https://github.com/databricks/spark-csv/pull/160#issuecomment-150200920, @HyukjinKwon posted a helpful suggestion on how to do this, modeled after the approach taken by ElasticSearch. I'm going to close this PR for now and will open a followup to try his approach.

PR #106, opened 2015-10-15
Edits for tutorial

JoshRosen:
This patch is a follow-up to #101 and makes many minor edits in the tutorial text.

/cc @sameeraxiomine


JoshRosen:
I'm going to go ahead and merge this now so that readers will find a more revised version of the tutorial. I can address any additional feedback in a followup commit.

PR #109, opened 2015-10-21
Fix bug related to AWS credentials in tempdir URI; fix typos in Hadoop config. instructions

### LINKED ISSUE #89: AWS config issues (python)
I've been working on getting my aws credentials properly set up to read/write, and encountered some issues:

1.) The given example has aws access being configured by sc.hadoopConfig.set("fs.s3n.awsAccessKeyId", "YOUR_KEY_ID"),etc ; there is no hadoopConfig method for the pyspark context.  It seems to be possible to pass a config file to hadoopFile, but this seems specific to whatever file you're pointing it at.

2.)  s3n://ACCESSKEY:SECRETKEY@bucket/path/to/temp/dir works absolutely fine when reading from redshift, but throws
<code>java.lang.IllegalArgumentException: AWS Access Key ID and Secret Access Key must be specified as the username or password (respectively) of a s3n URL</code>

when writing the dataframe back to redshift.

Both work fine if I define access/secret key environment variables before launching the shell, although the write throws 'invalid AVRO file found', probably because the column names contain _, as per https://github.com/databricks/spark-redshift/issues/84 .

JoshRosen:
This patch adds an integration test for writing and reading from Redshift while specifying AWS credentials via the `tempdir` URI instead of the HadoopConfiguration. See #89 for a discussion of one of these bugs.

I also corrected a bad typo in the Hadoop configuration section of the README: the old instructions used `sc.hadoopConf` instead of `sc.hadoopConfiguration.` I also added a section describing how to set this configuration from PySpark.

Fixes #89.


src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -136,13 +140,12 @@ private[redshift] case class RedshiftRelation(
...
-        unescaped.replace("'", "\\'")
-      }
-      s"SELECT $columnList FROM $tableNameOrSubquery $whereClause"
+      val escapedTableNameOrSubqury = tableNameOrSubquery.replace("'", "\\'")
+      s"SELECT $columnList FROM $escapedTableNameOrSubqury $whereClause"
     }
-    val fixedUrl = Utils.fixS3Url(tempDir)
+    // We need to remove S3 credentials from the unload path URI because they will conflict with
+    // the credentials passed via `credsString`.
+    val fixedUrl = Utils.fixS3Url(Utils.removeCredentialsFromURI(new URI(tempDir)).toString)
```
JoshRosen: If you don't do this, Redshift will complain. In the log message, Redshift will actually show the credentials in its error message, which it won't do if they're in `credsString`, so this also fixes a minor security problem.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -62,6 +65,8 @@ private[redshift] case class RedshiftRelation(
     }
   }
 
+  override def toString: String = s"RedshiftRelation($tableNameOrSubquery)"
```
JoshRosen: This change makes explain() output look a little nicer and also helps to prevent credentials from appearing in log messages.

src/test/scala/com/databricks/spark/redshift/UtilsSuite.scala
```diff
@@ -33,6 +33,12 @@ class UtilsSuite extends FunSuite with Matchers {
     Utils.joinUrls("s3n://foo/bar", "baz") shouldBe "s3n://foo/bar/baz/"
   }
 
+  test("joinUrls preserves credentials") {
+    assert(
+      Utils.joinUrls("s3n://ACCESSKEY:SECRETKEY@bucket/tempdir", "subdir") ===
```
JoshRosen: This is a regression test for a bug that was fixed in the last release.

marmbrus:
LGTM

JoshRosen:
Hmm, it looks like this test is failing in Travis but not locally. I'll try clearing my local environment to reproduce.

JoshRosen:
The problem was that our AWS secret key contained a slash. I've updated the Travis configs to fix this and everything is now working, so I'm going to merge this.

PR #110, opened 2015-10-21
Add example of writing data back to Redshift using SQL API

### LINKED ISSUE #81: Add example of writing data back to Redshift using SQL API
Add example of writing data back to Redshift using SQL API

JoshRosen:
This updates the README to include an example of saving data back to Redshift while using the SQL language API.

Fixes #81.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,13 +115,32 @@ df.write \
 
 #### SQL
 
+Reading data using SQL:
+
 ```sql
 CREATE TABLE my_table
 USING com.databricks.spark.redshift
-OPTIONS (dbtable 'my_table',
-         tempdir 's3n://my_bucket/tmp',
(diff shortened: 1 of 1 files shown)


JoshRosen:
After chatting with Michael, I realized that this technically supports "create if not exists" as well, so I'm going to update the new docs to correct this minor inaccuracy then will merge so we can release 0.5.2.

JoshRosen:
After some further investigation, I think that I'd like to leave this as-is. I don't think that the SQL API's "CREATE IF NOT EXISTS" corresponds to the SaveMode in the other language APIs, since it seems to check whether the _metastore_ table exists, not whether the Redshift table exists.

JoshRosen:
I'm going to merge this now so that we can release 0.5.2.

PR #114, opened 2015-10-28
Restore compatibility with spark-avro 1.0.0

### LINKED ISSUE #111: "Invalid operation: Mandatory url is not present in manifest file" error when writing data to Redshift while using an old version of spark-avro
This issue describes a problem that occurs when using `spark-redshift` with an old `spark-avro` dependency.

**Symptom**:

Saving data to Redshift fails with the following error if `spark-avro` 1.0.0 is used:

```scala
java.sql.SQLException: [Amazon](500310) Invalid operation: Mandatory url is not present in manifest file.
-----------------------------------------------
  error:  Mandatory url is not present in manifest file.
  code:      8001
  context:   Manifest file location=s3://spark-redshift-testing/temp/fc8db7a7-97fa-4dcd-b694-cbc44f8dbbe8/manifest.json url=s3://spark-redshift-testing/temp/fc8db7a7-97fa-4dcd-b694-cbc44f8dbbe8/part-r-00000-ro.avro
  query:     403423
  location:  s3_utility.cpp:328
  process:   padbmaster [pid=27093]
  -----------------------------------------------;
	at com.amazon.redshift.client.messages.inbound.ErrorResponse.toErrorException(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.handleErrorResponse(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getOperationMetadata(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getOperationMetadata(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.handleMessage(Unknown Source)
	at com.amazon.jdbc.communications.InboundMessagesPipeline.getNextMessageOfClass(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.doMoveToNextClass(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getReadyForQuery(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getOperationMetadata(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getOperationMetadata(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.handleMessage(Unknown Source)
	at com.amazon.jdbc.communications.InboundMessagesPipeline.getNextMessageOfClass(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.doMoveToNextClass(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.doMoveToNextClass(Unknown Source)
	at com.amazon.redshift.client.PGMessagingContext.getErrorResponse(Unknown Source)
	at com.amazon.redshift.client.PGClient.executePreparedStatement(Unknown Source)
	at com.amazon.redshift.dataengine.PGIQueryExecutor.execute(Unknown Source)
	at com.amazon.jdbc.common.SPreparedStatement.executeWithParams(Unknown Source)
	at com.amazon.jdbc.common.SPreparedStatement.execute(Unknown Source)
	at com.databricks.spark.redshift.RedshiftWriter$$anonfun$com$databricks$spark$redshift$RedshiftWriter$$doRedshiftLoad$1.apply(RedshiftWriter.scala:170)
	at com.databricks.spark.redshift.RedshiftWriter$$anonfun$com$databricks$spark$redshift$RedshiftWriter$$doRedshiftLoad$1.apply(RedshiftWriter.scala:165)
	at scala.Option.foreach(Option.scala:236)
	at com.databricks.spark.redshift.RedshiftWriter.com$databricks$spark$redshift$RedshiftWriter$$doRedshiftLoad(RedshiftWriter.scala:165)
	at com.databricks.spark.redshift.RedshiftWriter.saveToRedshift(RedshiftWriter.scala:375)
	at com.databricks.spark.redshift.DefaultSource.createRelation(DefaultSource.scala:106)
	at org.apache.spark.sql.sources.ResolvedDataSource$.apply(ddl.scala:309)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:144)
```

**Cause**:

In order to fix a crash that could occur when writing tables containing empty partitions, #99 modified the write path to use manifest files that instruct Redshift to load only the non-empty partitions' Avro files. The code that handled processing of filenames when generating the manifest made assumptions that hold for `spark-avro` 2.0.0+ but not for `spark-avro` 1.0.0.

**Solution**:

For now, the best solution is to user `spark-avro` 2.0.1. Usually this should happen automatically if you install libraries using Maven or another package/dependency manager.

~~Databricks users can attach `spark-redshift` by specifying the coordinate `com.databricks:spark-redshift_2.10:0.5.2` in the Maven library upload screen or by using the integrated Spark Packages and Maven Central browser). If you are manually attaching JARs to your cluster, you can either include `spark-avro` 2.0.1+ or you can generate a combined assembly JAR containing all of `spark-redshift`'s dependencies by using the `sbt-assembly` plugin: just run `sbt/sbt assembly` in the `spark-redshift` checkout.~~ *EDIT: * the approach listed here will not work for DBC 1.4.1 because the driver's own `spark-avro` dependency will take precedence. We'll have to either shade or restore compatibility.

In the long term, I may be able to fix this by either restoring compatibility with `spark-avro` 1.0.0 or by adding an error message so we fail in a less confusing way when an incompatible version of the library is used.

JoshRosen:
This patch addresses an issue where `spark-redshift` would run into "Mandatory url is not present in manifest file" errors when saving data back to Redshift if a pre-2.0.0 version of `spark-avro` was used at runtime. This problem should not arise for most users of `spark-redshift`, since the proper version of `spark-avro` should automatically be pulled in via Maven or Ivy; the goal of this patch is to provide compatibility for users who cannot upgrade `spark-avro` for other compatibility-related reasons.

**Cause of the write path bug**: In order to fix a crash that could occur when writing tables containing empty partitions, #99 modified the write path to use manifest files that instruct Redshift to load only the non-empty partitions' Avro files. The code that handled processing of filenames when generating the manifest made assumptions about filenames that hold for spark-avro 2.0.0+ but not for spark-avro 1.0.0.

**Solution**: use a different method to build the correct list of part file names when constructing the manifest.

Fixes #111.


.travis.yml
```diff
@@ -13,20 +13,26 @@ matrix:
...
     - jdk: openjdk7
       scala: 2.10.5
-      env: HADOOP_VERSION="2.2.0" SPARK_VERSION="1.5.0"
+      env: HADOOP_VERSION="2.2.0" SPARK_VERSION="1.5.0" SPARK_AVRO_VERSION="2.0.1"
+    # Configuration corresponding to DBC 1.4.x driver package as of DBC 2.4,
+    # which uses spark-avro 1.0.0. We use Hadoop 2.2.0 here, while DBC uses
+    # 1.2.1, because the 1.4.1 published to Maven Central is a Hadoop 2.x build.
+    - jdk: openjdk7
+      scala: 2.10.5
+      env: HADOOP_VERSION="2.2.0" SPARK_VERSION="1.4.1" SPARK_AVRO_VERSION="1.0.0"
```
JoshRosen: This configuration acts as a regression test for this issue.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -311,29 +311,25 @@ private[redshift] class RedshiftWriter(
...
-      // The saved filenames are going to be of the form part-r-XXXXX-UUID.avro. There isn't a way
-      // to figure out what this UUID is, besides querying S3. In principle, this file format could
-      // change in a way that breaks this library, but this seems unlikely and we can patch around
-      // it if that happens.
+      // The saved filenames depend on the spark-avro version. In spark-avro 1.0.0, the write
+      // path uses SparkContext.saveAsHadoopFile(), which produces filenames of the form
+      // part-XXXXX.avro. In spark-avro 2.0.0+, the partition filenames are of the form
+      // part-r-XXXXX-UUID.avro.
       val fs = FileSystem.get(URI.create(tempDir), sqlContext.sparkContext.hadoopConfiguration)
-      val writeJobUUID = fs.listStatus(new Path(tempDir))
```
JoshRosen: In terms of efficiency: note that here we call `fs.listStatus()` on the root of the write-specific `tempDir` and we make the same `listStatus` call in the updated code below; the only change is that we now filter this list rather than trying to extract the UUID and use that to construct the right filenames.

Both versions of the code potentially run into problems when listing directories with millions of files; for this reason, I'm wondering whether I should use the Amazon S3 client to perform this listing rather than using the FileSystem API.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -311,29 +311,25 @@ private[redshift] class RedshiftWriter(
...
-        .iterator
-        .map(_.getPath.getName)
-        .filter(_.startsWith("part"))
-        .take(1)
-        .toSeq
-        .headOption.getOrElse(throw new Exception("No part files were written!"))
-        .drop("part-r-XXXXX-".length)
-        .stripSuffix(".avro")
-
+      val partitionIdRegex = "^part-(?:r-)?(\\d{5}).*$".r
```
JoshRosen: It just occurred to me: what if you have more than 99999 partitions? What format does the filename have in that case?

JoshRosen: I'll update this to not hardcode the 5-digit requirement.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
JoshRosen: Actually, this shouldn't be a problem: if you have _millions_ of partitions then it almost certainly makes more sense to coalesce them prior to writing to Redshift. I can add this auto-coalescing in a followup.

PR #117, opened 2015-11-01
Issue JDBC queries asynchronously in order to support cancellation / InterruptedException

### LINKED ISSUE #116: Cancel Redshift queries when thread is interrupted
It appears that blocking JDBC calls do not respond to thread interrupts, preventing us from canceling reads or writes. It appears that the easiest solution is to issue the JDBC statements from a separate thread or Future, then use `Statement.cancel()` to stop the query (see http://blog.grovehillsoftware.com/2009/08/interruptible-jdbc-statements.html). The easiest way to do this is probably to define a utility method in our JDBCWrapper that encapsulates this logic.

JoshRosen:
Most blocking JDBC calls do not respond to InterruptedException, which causes problems when users try to cancel long-running `spark-redshift` loads or unloads. This patch addresses this issue by using issuing the blocking JDBC calls in Futures.

Fixes #116.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -41,15 +41,16 @@ private[redshift] class JDBCWrapper {
 
   private val ec: ExecutionContext = {
     val threadFactory = new ThreadFactory {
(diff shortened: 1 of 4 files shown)


JoshRosen:
Tested this out in a Databricks notebook and confirmed that cancellation now works as expected (doesn't crash REPL and successfully terminates the query in Redshift).

Issue #118, opened 2015-11-02
StringType defaults to TEXT which is varchar(256)


eprochasson:
Hi,

My understanding is that spark-redshift takes a DataFrame of row, with its schema, and cast it to the proper schema for Redshift. `StringType` becomes `TEXT`, which is understood by Redshift as `varchar(256)`, hence fails if the text fields contains a string longer than 256 bytes.

It works in append mode if the table has been created before (otherwise I think it goes through a temp table, using the automatically inferred schema, and fail).

JoshRosen:
Hi @eprochasson,

This is a known behavior; see #29. We have (partially) addressed this by allowing the size of individual string columns via Spark SQL's column metadata feature: https://github.com/databricks/spark-redshift#configuring-the-maximum-size-of-string-columns. However, this approach's ease-of-use is limited by Spark SQL's limited support for column metadata configuration. I'd love to have your feedback on whether our existing column-metadata-based approach addresses your needs or whether you have any suggestions for how we can make this easier to configure (or better-documented).

eprochasson:
I'll look into it. Sorry I missed it!

eprochasson:
It works beautifully. Minor issue in the documentation is that `df.withColumn` returns a new DataFrame and does not modify the existing one (so that snippet won't work).

I wrote a short recipe for it here https://gist.github.com/eprochasson/d1feb24eb36eb882d580

You're very welcome to integrate it to the README if you feel it's worth it.

Thanks a ton.

JoshRosen:
If the current README instructions don't work as written then we should update them. Want to submit a PR and I'll review this week?

eprochasson:
Indeed it doesn't (it's actually not even syntactically correct, it's missing a parenthesis). I'll submit a PR as soon as I have a minute.

robbyki:
Apologies for misunderstanding this issue but I'm going through several resources to try and understand how to maintain my schema created outside of spark and then just truncating my tables from spark followed by writing with a savemode of overwrite. My problem exactly this issue with respect to my db netezza failing when it sees spark trying to save a text data type so I then have to go specify in my new jdbc dialect to use varchar(n) which does work however that just replaces all of my varchar columns (different lengths for different columns) with whatever I specified in my dialect which is not what I want. How can I just have it save the TEXT as varchar without specifying a length in the custom dialect?

yobdoy:
I was having this same problem, this [answer][1] worked for me:

[...]

PR #119, opened 2015-11-05
Fix for #118: StringType defaults to TEXT which is varchar(256)

eprochasson:
also fixed a bit of documentation about compound key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,11 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to 
 
 Here is an example of updating a column's metadata field in Scala:
 
-```
+```scala
 import org.apache.spark.sql.types.MetadataBuilder
-val metadata = new MetadataBuilder().putLong("maxlength", 10).build()
-df.withColumn("colName", col("colName").as("colName", metadata)
-```
+
+    // Specify the custom width of each column
+    val columnWidthMap = Map(
+      "column1" -> 10,
+      "column2" -> 11,
+      "column3" -> 12,
+      "column4" -> 13,
+      "column5" -> 14,
+    )
+
+    var df // the dataframe you'll want to write to redshift
+
+    // Apply each column metadata customization
+    columnWidthMap.foreach(m => {
+      df = df.withColumn(m._1, df(m._1).as(m._1, new MetadataBuilder().putLong("maxlength", m._2).build()))
+    })
+
+    df.write
+      .format("com.databricks.spark.redshift")
+      .option("url", jdbcURL)
+      .option("tempdir", s3TempDirectory)
+      .option("dbtable", sessionTable)
+      .mode(SaveMode.Overwrite)
+      .save()```
 
 Column metadata modification is unsupported in the Python, SQL, and R language APIs.
 
--- a/README.md
+++ b/README.md
@@ -350,33 +350,33 @@ When creating Redshift tables, `spark-redshift`'s default behavior is to create 
 
 To support larger columns, you can use the `maxlength` column metadata field to specify the maximum length of individual string columns. This can also be done as a space-savings performance optimization in order to declare columns with a smaller maximum length than the default.
 
-Here is an example of updating a column's metadata field in Scala:
+Here is an example of updating multiple columns' metadata fields:
 
 ```scala
 import org.apache.spark.sql.types.MetadataBuilder
 
 // Specify the custom width of each column
-val columnWidthMap = Map(
-  "column1" -> 10,
-  "column2" -> 11,
-  "column3" -> 12,
-  "column4" -> 13,
-  "column5" -> 14,
+val columnLengthMap = Map(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -352,10 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to
...
+columnWidthMap.foreach(m => {
+  df = df.withColumn(m._1, df(m._1).as(m._1, new MetadataBuilder().putLong("maxlength", m._2).build()))
+})
+
+df.write
+  .format("com.databricks.spark.redshift")
+  .option("url", jdbcURL)
+  .option("tempdir", s3TempDirectory)
+  .option("dbtable", sessionTable)
+  .mode(SaveMode.Overwrite)
```
JoshRosen: Suuuuuper minor nit, but I'm slightly wary of putting this into example code that a user might verbatim copy-and-paste. Probably better to omit the use of a more powerful SaveMode.

README.md
```diff
@@ -352,10 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to
...
+  "column2" -> 11,
+  "column3" -> 12,
+  "column4" -> 13,
+  "column5" -> 14,
+)
+
+var df // the dataframe you'll want to write to redshift
+
+// Apply each column metadata customization
+columnWidthMap.foreach(m => {
```
JoshRosen: Style nit: you could write `columnWidthMap.foreach { m =>` in order to omit an extra layer of braces.

README.md
```diff
@@ -352,10 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to
...
+  df = df.withColumn(m._1, df(m._1).as(m._1, new MetadataBuilder().putLong("maxlength", m._2).build()))
```
JoshRosen: This line is kinda long, so you could split it into two by defining `val metadata = ...` on the previous line, then using that metadata when re-assigning to `df`.

README.md
```diff
@@ -352,10 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to
...
+// Specify the custom width of each column
+val columnWidthMap = Map(
+  "column1" -> 10,
+  "column2" -> 11,
+  "column3" -> 12,
+  "column4" -> 13,
+  "column5" -> 14,
+)
+
+var df // the dataframe you'll want to write to redshift
```
JoshRosen: Nit: capitalize Redshift whenever it appears, since it's the name of a product.

README.md
```diff
@@ -352,10 +352,32 @@ To support larger columns, you can use the `maxlength` column metadata field to
 
 Here is an example of updating a column's metadata field in Scala:
```
JoshRosen: "a column's metadata field" -> "multiple columns' metadata fields"?

README.md
JoshRosen: Also, could write `columnWidthMap.foreach { case (colName, length) =>` to avoid the `_1`s.

README.md
```diff
@@ -294,9 +294,9 @@ must also set a distribution key with the <tt>distkey</tt> option.
 
 <p>Examples include:</p>
 <ul>
-    <li><tt>SORTKEY my_sort_column</tt></li>
-    <li><tt>COMPOUND SORTKEY sort_col_1, sort_col_2</tt></li>
-    <li><tt>INTERLEAVED SORTKEY sort_col_1, sort_col_2</tt></li>
+    <li><tt>SORTKEY(my_sort_column)</tt></li>
```
JoshRosen: Good catch (I never tried this feature myself since it predates my work on this project).

JoshRosen:
Thanks for submitting this PR. I left a bunch of nitpicky comments because I'm kind of a perfectionist; if you don't have time to address them, I can take care of them myself on merge tomorrow.

eprochasson:
Ha! Neh I'll do it.

eprochasson:
I made the example slightly less abstract too.

README.md
```diff
@@ -350,12 +350,34 @@ When creating Redshift tables, `spark-redshift`'s default behavior is to create
...
+  "language_code" -> 2,
+  "country_code" -> 2,
+  "url" -> 2083
+)
+
+var df = ... // the dataframe you'll want to write to Redshift
+
+// Apply each column metadata customization
+columnLengthMap.foreach(
+  case (colName, length) => {
```
JoshRosen: You can still remove one level of braces, but I'll do this myself right now on merge.

JoshRosen:
Merged. Thanks!

README.md
```diff
@@ -350,12 +350,34 @@ When creating Redshift tables, `spark-redshift`'s default behavior is to create
...
+    val metadata = new MetadataBuilder().putLong("maxlength", length).build())
```
JoshRosen: Noticed that there was also an extra `)` here, so I fixed that as well.

PR #120, opened 2015-11-09
Fix instance profile config property name in docs

jacek-rzrz:
temporary_aws_access_key_id was listed twice and
temporary_aws_secret_access_key was missing.


JoshRosen:
Good catch; thanks for fixing this. I'm going to merge this now.

PR #127, opened 2015-11-23
Re-use Row instance itself in row conversion function

JoshRosen:
As a minor performance optimization, we can re-use the same `Row` instance itself in the row converter function.


JoshRosen:
Going to close this for now; will revisit once we can depend on nicer stable APIs for this.

PR #128, opened 2015-11-24
Implement Spark 1.6's new unhandledFilters API

JoshRosen:
Spark 1.6 extends `BaseRelation` with a new API which allows data sources to tell Spark which filters they handle, allowing Spark to eliminate its own defensive filtering for filters that are handled by the data source (see https://github.com/apache/spark/pull/9399 for more details).

This patch implements this new API in `spark-redshift` and adds tests.


rxin:
LGTM

Issue #122, opened 2015-11-17
DATEFORMAT issue in RedshiftWriter.scala


jackar:
There's no need for a dateformat - especially since dates don't have millisecond level granularity, that would be timeformat, but that's already the default...

http://docs.aws.amazon.com/redshift/latest/dg/r_DATEFORMAT_and_TIMEFORMAT_strings.html

````
private def copySql(
      sqlContext: SQLContext,
      params: MergedParameters,
      creds: AWSCredentials,
      manifestUrl: String): String = {
    val credsString: String = AWSCredentialsUtils.getRedshiftCredentialsString(creds)
    val fixedUrl = Utils.fixS3Url(manifestUrl)
    s"COPY ${params.table.get} FROM '$fixedUrl' CREDENTIALS '$credsString' FORMAT AS " +
      s"AVRO 'auto' DATEFORMAT 'YYYY-MM-DD HH:MI:SS' manifest ${params.extraCopyOptions}"
  }
````

"DATEFORMAT 'YYYY-MM-DD HH:MI:SS' " should be removed from Line 101

I found this issue when reading a parquet file from s3 and then trying to save the dataframe into redshift:

````
Error (code 1205) while loading data into Redshift: "Date value did not match format specified [YYYY-MM-DD HH:MI:SS]"
````

The string format for date I'm using is YYYY-MM-DD

JoshRosen:
Hi @jackar,

To provide a bit of historical context: `spark-avro` does not support date or timestamp columns, so in order to generate the unload data we end up converting dates and timestamps into strings prior to loading them into Redshift. See #39 for a discussion of why we use this particular string format, vs. other formats.

To help me understand the error message that you saw, can you share the schema of the DataFrame that you were saving back to Redshift? If you have a string column in YYYY-MM-DD format and are relying on Redshift to perform the conversion, then I don't think that's going to be compatible with the formats that we need to set in order to handle timestamp conversion. In that case, my recommendation would be to use Spark's built-in date and time functions to convert the YYYY-MM-DD  into a SQL date prior to saving to Redshift.

JoshRosen:
According to the [Redshift documentation](https://docs.aws.amazon.com/redshift/latest/dg/r_Examples_with_datetime_types.html):

> If you attempt to insert a timestamp value into a DATE column, the time literal is truncated and the date itself is loaded.

[...]

PR #130, opened 2015-11-25
Change RedshiftWriter Dateformat to yyyy-MM-dd

jackar:
This patch modifies the write path to write dates in `yyyy-MM-dd` format instead of the timestamp format. See #122 for more details.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Conversions.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Conversions.scala
@@ -148,9 +148,9 @@ private[redshift] class RedshiftDateFormat extends DateFormat {
   private val redshiftDateFormat = new SimpleDateFormat(PATTERN)
 
   override def format(
-                       date: Date,
-                       toAppendTo: StringBuffer,
-                       fieldPosition: FieldPosition): StringBuffer = {
+     date: Date,
+     toAppendTo: StringBuffer,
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/redshift/Conversions.scala
```diff
@@ -138,3 +138,23 @@ private[redshift] class RedshiftTimestampFormat extends DateFormat {
...
+private[redshift] class RedshiftDateFormat extends DateFormat {
+
+  // Imports and exports with Redshift require that dates are represented
+  // as strings, using the following format
+  private val PATTERN = "yyyy-MM-dd"
+
+  private val redshiftDateFormat = new SimpleDateFormat(PATTERN)
+
+  override def format(
+                       date: Date,
```
JoshRosen: Minor styling nit: can you indent the parameters by only four spaces relative to the start of the previous line in order to match the style on line 126 in this same file?

JoshRosen:
Thanks for the PR. I'm going to push your change to a temporary branch so that the integration tests can run against a real Redshift cluster (for security reasons, Travis won't run the credential-injection for public pull request builds).

JoshRosen:
This passed full integration tests (https://travis-ci.org/databricks/spark-redshift/builds/93199052) and looks good to me, so I'm going to merge it now. Thanks@

PR #141, opened 2015-12-27
Minor fixes to unit test mocks in order to suppress spurious exceptions

JoshRosen:
This patch contains two minor bug fixes related to unit test mocks:

- In `RedshiftSourceSuite`, `mockS3Client` needs to be created in `beforeEach` instead of in `beforeAll`; the old behavior was causing NullPointerExceptions in tests because the `afterEach` method would teardown the `mockS3Client` but it would never be re-created for later tests.
- In `MockRedshift`, we need to stub both `execute()` _and_ `executeQuery()` when mocking a `PreparedStatement`; this is necessary after the changes in #117.

### DIFF
--- a/src/test/scala/com/databricks/spark/redshift/MockRedshift.scala
+++ b/src/test/scala/com/databricks/spark/redshift/MockRedshift.scala
@@ -16,7 +16,7 @@
 
 package com.databricks.spark.redshift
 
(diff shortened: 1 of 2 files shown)


JoshRosen:
Looks like the lone test failure is a test that's known to be slightly flaky in Travis (I'll investigate separately). Each test passed at least once in one of the builders, so I'm going to merge this now.

Issue #138, opened 2015-12-24
Warning: An error occurred while trying to read the S3 bucket lifecycle configuration


wangchuan3533:
```
15/12/24 10:46:47 WARN Utils$: An error occurred while trying to read the S3 bucket lifecycle configuration
java.lang.NullPointerException
	at com.databricks.spark.redshift.Utils$.checkThatBucketHasObjectLifecycleConfiguration(Utils.scala:94)
	at com.databricks.spark.redshift.RedshiftWriter.saveToRedshift(RedshiftWriter.scala:368)
	at com.databricks.spark.redshift.DefaultSource.createRelation(DefaultSource.scala:106)
	at org.apache.spark.sql.execution.datasources.ResolvedDataSource$.apply(ResolvedDataSource.scala:170)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:146)
	at $line22.$read$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC.<init>(<console>:22)
	at $line22.$read$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC.<init>(<console>:27)
	at $line22.$read$$iwC$$iwC$$iwC$$iwC$$iwC$$iwC.<init>(<console>:29)
	at $line22.$read$$iwC$$iwC$$iwC$$iwC$$iwC.<init>(<console>:31)
	at $line22.$read$$iwC$$iwC$$iwC$$iwC.<init>(<console>:33)
	at $line22.$read$$iwC$$iwC$$iwC.<init>(<console>:35)
	at $line22.$read$$iwC$$iwC.<init>(<console>:37)
	at $line22.$read$$iwC.<init>(<console>:39)
	at $line22.$read.<init>(<console>:41)
	at $line22.$read$.<init>(<console>:45)
	at $line22.$read$.<clinit>(<console>)
	at $line22.$eval$.<init>(<console>:7)
	at $line22.$eval$.<clinit>(<console>)
	at $line22.$eval.$print(<console>)
```

JoshRosen:
[...]

PR #142, opened 2015-12-27
Handle non-existent bucket in bucket lifecycle check

JoshRosen:
This patch improves the error-handling in `Utils.checkThatBucketHasObjectLifecycleConfiguration` so that non-existent buckets result in a proper error message rather than a confusing NullPointerException (see #138).


JoshRosen:
Going to merge this now.

PR #143, opened 2015-12-28
Fix bugs related to JDBC driver precedence and PostgreSQL JDBC subprotocol

### LINKED ISSUE #126: java.lang.IllegalArgumentException: Unsupported JDBC protocol: 'postgresql'
How to use postgresql jdbc driver with spark-redshift? 
following code give me exception 
java.lang.IllegalArgumentException: Unsupported JDBC protocol: 'postgresql'

```scala
val df1: DataFrame = sqlContext.read
			.option("url", "jdbc:postgresql://host:5439/db?user=test&password=test")
			.option("dbtable", "wdata")
			.option("tempdir", "s3n://accessKEy:SecretKEy@redshift/dir/")
			.load()
		df1.show()
```

Full stacktrace is as follow

```
Exception in thread "main" java.lang.IllegalArgumentException: Unsupported JDBC protocol: 'postgresql'
	at com.databricks.spark.redshift.JDBCWrapper$$anonfun$getDriverClass$2.apply(RedshiftJDBCWrapper.scala:68)
	at com.databricks.spark.redshift.JDBCWrapper$$anonfun$getDriverClass$2.apply(RedshiftJDBCWrapper.scala:52)
	at scala.Option.getOrElse(Option.scala:120)
	at com.databricks.spark.redshift.JDBCWrapper.getDriverClass(RedshiftJDBCWrapper.scala:51)
	at com.databricks.spark.redshift.JDBCWrapper.getConnector(RedshiftJDBCWrapper.scala:138)
	at com.databricks.spark.redshift.RedshiftRelation$$anonfun$schema$1.apply(RedshiftRelation.scala:59)
	at com.databricks.spark.redshift.RedshiftRelation$$anonfun$schema$1.apply(RedshiftRelation.scala:56)
	at scala.Option.getOrElse(Option.scala:120)
	at com.databricks.spark.redshift.RedshiftRelation.schema$lzycompute(RedshiftRelation.scala:56)
	at com.databricks.spark.redshift.RedshiftRelation.schema(RedshiftRelation.scala:55)
	at org.apache.spark.sql.execution.datasources.LogicalRelation.<init>(LogicalRelation.scala:31)
	at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:120)
```

JoshRosen:
This patch fixes two bugs related to JDBC drivers:

- The `RedshiftJDBCWrapper.getDriverClass` method mistakenly used `postgres` as the PostgreSQL JDBC subprotocol, whereas the correct subprotocol is actually `postgresql` (see #126). This was straightforward to fix.
- When the user specified an explicit JDBC driver class, we would register that driver with the JDBC DriverManager but would not necessarily use it to create connections. In a nutshell, the problem is that you might have multiple JDBC drivers on your classpath that claim to be able to handle the same subprotocol and there doesn't seem to be an intuitive way to control which of those drivers takes precedence. The change implemented in this patch is to iterate over the driver registry's loaded drivers in order to obtain the correct driver and then use that driver to create a connection.

Fixes #126.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -17,12 +17,12 @@
 
 package com.databricks.spark.redshift
 
-import java.net.URI
 import java.sql.{ResultSet, PreparedStatement, Connection, Driver, DriverManager, ResultSetMetaData, SQLException}
 import java.util.Properties
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.{ThreadFactory, Executors}
 
+import scala.collection.JavaConverters._
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration.Duration
 import scala.util.Try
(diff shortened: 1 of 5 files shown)


JoshRosen:
@yhuai @marmbrus, the JDBC driver precedence issue is not necessarily a bug given our user-facing documentation (if you carefully read what the documentation actually promises), but the current behavior is nonintuitive to me and I think that the new behavior implemented here is easier to reason about and more in line with what users might expect. I'd like to discuss whether we should make a similar change in Spark's JDBC data source, since I think that the new behavior implemented in this patch would allow us to work around some of the other JDBC driver precedence issues that we've seen in Spark (such as the issue with the Redshift driver taking precedence over the Postgres one for `jdbc:postgresql/` URIs).

This change was motivated by the difficulties that I had while trying to write the Postgres integration suite that's included in this patch.

yhuai:
looks good to me

JoshRosen:
Cool. Let me take a look at the README / code comments to make sure that they reflect this change, then I'll merge.

I'm going to open a PR against Spark so that we mirror the same behavior there.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -194,9 +199,19 @@ private[redshift] class JDBCWrapper {
...
-    DriverManager.getConnection(url, new Properties())
+    val driverClass: String = getDriverClass(subprotocol, userProvidedDriverClass)
+    registerDriver(driverClass)
+    // Note that we purposely don't call DriverManager.getConnection() here: we want to ensure
+    // that an explicitly-specified user-provided driver class can take precedence over the default
+    // class, but DriverManager.getConnection() might return a according to a different precedence.
+    // At the same time, we don't want to create a driver-per-connection, so we use the
+    // DriverManager's driver instances to handle that singleton logic for us.
+    val driver: Driver = DriverManager.getDrivers.asScala.collectFirst {
+      case d if d.getClass.getCanonicalName == driverClass => d
```
JoshRosen: I just realized that there's a bit of a bug here: if `d` is an instance of `DriverWrapper` then we need to compare against `d.wrapped.getClass.getCanonicalName`. I'll open a followup PR to fix this.

PR #147, opened 2016-01-04
Handle DriverWrapper when scanning registered drivers in JDBWrapper.getConnector()

JoshRosen:
This is a followup to #143 which fixes a corner-case bug in our scanning of registered JDBC drivers: we need to properly handle Spark's `DriverWrapper` drivers, which are used to wrap JDBC drivers in order to make them accessible from the root classloader so that the `DriverManager` can find them.

A simpler, reflection-free version of this change was incorporated into https://github.com/apache/spark/pull/10519

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -201,12 +201,23 @@ private[redshift] class JDBCWrapper {
     val subprotocol = url.stripPrefix("jdbc:").split(":")(0)
     val driverClass: String = getDriverClass(subprotocol, userProvidedDriverClass)
     registerDriver(driverClass)
+    val driverWrapperClass: Class[_] = if (SPARK_VERSION.startsWith("1.4")) {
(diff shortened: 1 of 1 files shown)


JoshRosen:
Unfortunately, this is a little tricky to test in an automated fashion because we need to rig the classloaders in order to make sure that the test JDBC drivers aren't loaded by the root classloader.

As a pragmatic trade-off, I'm inclined to test this manually using `spark-shell` and `--packages` with the Postgres driver. I'll do this tomorrow.

JoshRosen:
I tested this manually with `spark-shell`, `--jars`, and a modified copy of this code with additional logging statements. I ran my manual tests on Spark 1.4.1, 1.5.1, and 2.0.0-SNAPSHOT, so I'm confident that the portions of the reflection code here which aren't covered by unit tests will work in those versions.

In case we don't get automated tests for this in place by the time of the next release, I'll open a PR to document the manual testing procedure and will create a checklist so future release coordinators / maintainers can know how to ensure that this doesn't break.

I'm going to merge this now and will begin preparation for a new `spark-redshift` release tomorrow.

PR #150, opened 2016-01-12
Add documentation on transactional guarantees

JoshRosen:
This pull request adds documentation on `spark-redshift`'s transactional guarantees.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ This section describes `spark-redshift`'s transactional guarantees. For general 
 
 **Appending to an existing table**: Writes use Redshift's [`COPY`](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command, which is [atomic and transactional](https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-multiple-files.html). When loading data into Redshift, `spark-redshift` uses [manifests](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) to guard against S3 eventual-consistency. Thus, `spark-redshift` appends to existing tables have the same atomic and transactional properties as regular Redshift `COPY` commands.
 
-**Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` suceeeds, it will use a transaction to atomically delete the overwritten table and rename the temporay table to destination table.
+**Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` succeeds, it will use a transaction to atomically delete the overwritten table and rename the temporary table to destination table.
 
 This behavior can be disabled by setting `usestagingtable` to `false`, in which case the destination table will be deleted before the `COPY`.
 
--- a/README.md
+++ b/README.md
@@ -380,6 +380,20 @@ df.write
   .save()
 ```
 
+## Transactional Guarantees
+
+This section describes `spark-redshift`'s transactional guarantees. For general information on Redshift's transactional guarantees, see the [Managing Concurrent Write Operations](https://docs.aws.amazon.com/redshift/latest/dg/c_Concurrent_writes.html) chapter in the Redshift documentation.
+
+**Appending to an existing table**: Writes use Redshift's [`COPY`](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command, which is [atomic and transactional](https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-multiple-files.html). When loading data into Redshift, `spark-redshift` uses [manifests](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) to guard against S3 eventual-consistency. Thus, `spark-redshift` appends to existing tables have the same atomic and transactional properties as regular Redshift `COPY` commands.
+
+**Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` suceeeds, it will use a transaction to atomically delete the overwritten table and rename the temporay table to destination table.
+
+This behavior can be disabled by setting `usestagingtable` to `false`, in which case the destination table will be deleted before the `COPY`.
+
+**Creating a new table**: `spark-redshift` currently does not use a transaction when creating a new table, so the creation of the empty table and the initial `COPY` to that table may become visible at different times to readers. The `COPY` itself is atomic, so the table will never be visible in a state where it contains a non-empty subset of the saved rows.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -380,6 +381,22 @@ df.write
...
+
+**Reading**: Reads use Redshift's [`UNLOAD`](https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) command. According to the documentation for Redshift's [`BEGIN`](https://docs.aws.amazon.com/redshift/latest/dg/r_BEGIN.html) command, "[although] you can use any of the four transaction isolation levels, Amazon Redshift processes all isolation levels as serializable." `UNLOAD` saves the result of a `SELECT` query and this `SELECT` query should observe a consistent snapshot of the database.
+
+According to the [Amazon S3 Data Consistency Model](https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel) documentation, S3 bucket listing is eventually-consistent. In `spark-redshift` 1.6.0 and earlier, the S3 read path performs bucket-listing and thus may be impacted by this eventual-consistency, meaning that in rare circumstances reads may see only a subset of the unloaded data. This will be fixed in future releases by using `UNLOAD`'s `MANIFEST` support.
```
JoshRosen: Since I'm going to fix this in https://github.com/databricks/spark-redshift/pull/151, I should probably just update this now to move the current text into a pre-1.6.1 caveat / callout.

README.md
```diff
@@ -380,6 +381,22 @@ df.write
...
+
+**Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` succeeds, it will use a transaction to atomically delete the overwritten table and rename the temporary table to destination table.
+
+This behavior can be disabled by setting `usestagingtable` to `false`, in which case the destination table will be deleted before the `COPY`.
+
+**Creating a new table**: `spark-redshift` currently does not use a transaction when creating a new table, so the creation of the empty table and the initial `COPY` to that table may become visible at different times to readers. The `COPY` itself is atomic, so the table will never be visible in a state where it contains a non-empty subset of the saved rows.
```
JoshRosen: Let me double-check to make sure this is actually true. If it's actually true that creating a new table isn't atomic, then it should be trivially easy to fix.

README.md
```diff
@@ -380,6 +381,22 @@ df.write
   .save()
 ```
 
+## Transactional Guarantees
+
+This section describes `spark-redshift`'s transactional guarantees. For general information on Redshift's transactional guarantees, see the [Managing Concurrent Write Operations](https://docs.aws.amazon.com/redshift/latest/dg/c_Concurrent_writes.html) chapter in the Redshift documentation.
+
+**Appending to an existing table**: Writes use Redshift's [`COPY`](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command, which is [atomic and transactional](https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-multiple-files.html). When loading data into Redshift, `spark-redshift` uses [manifests](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) to guard against S3 eventual-consistency. Thus, `spark-redshift` appends to existing tables have the same atomic and transactional properties as regular Redshift `COPY` commands.
```
yhuai: Maybe we should be more specific on `Writes` that we mentioned in `Writes use Redshift's COPY...`? What are those `Writes`? 

Also, do we need to explain why `manifests` can guard against the eventual-consistency issue?

README.md
```diff
@@ -380,6 +381,22 @@ df.write
...
+
+**Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` succeeds, it will use a transaction to atomically delete the overwritten table and rename the temporary table to destination table.
```
yhuai: Am I right that temp table creation and appending data to that using copy statement do not use transaction?

yhuai: ah, i see. For creating the temp table, the semantic is following `Creating a new table`, right?

README.md
```diff
@@ -380,6 +381,22 @@ df.write
...
+
+This behavior can be disabled by setting `usestagingtable` to `false`, in which case the destination table will be deleted before the `COPY`.
+
+**Creating a new table**: `spark-redshift` currently does not use a transaction when creating a new table, so the creation of the empty table and the initial `COPY` to that table may become visible at different times to readers. The `COPY` itself is atomic, so the table will never be visible in a state where it contains a non-empty subset of the saved rows.
+
+**Reading**: Reads use Redshift's [`UNLOAD`](https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) command. According to the documentation for Redshift's [`BEGIN`](https://docs.aws.amazon.com/redshift/latest/dg/r_BEGIN.html) command, "[although] you can use any of the four transaction isolation levels, Amazon Redshift processes all isolation levels as serializable." `UNLOAD` saves the result of a `SELECT` query and this `SELECT` query should observe a consistent snapshot of the database.
```
yhuai: What is the connection between `UNLOAD` and `BEGIN`?

README.md
yhuai: Maybe we need to more specific on when `eventual-consistency` may have a impact because `Amazon S3 buckets in all Regions provide read-after-write consistency for PUTS of new objects and eventual consistency for overwrite PUTS and DELETES. ` (https://aws.amazon.com/s3/faqs/)?

README.md
JoshRosen: > Am I right that temp table creation and appending data to that using copy statement do not use transaction?

That's correct, although for simplicity / ease-of-reasoning I'm thinking that we should create the temp table inside of a transaction to ensure that it's deleted properly in case the load is interrupted. I'm wondering whether I should just do that cleanup ahead of the documentation patch here.

README.md
JoshRosen: By writes, I mean any operation which insert rows into a Redshift table.

I don't think we need to explain the manifests in any more detail than has been done here, since interested users can just click through to read that link.

README.md
JoshRosen: Yeah, this was a subtlety that I didn't go into here. The TL;DR of it is that we don't actually issue the UNLOAD inside of a begin / end pair, but I don't think that makes a difference since UNLOAD itself is treated as a single atomic operation. My reading of Redshift's isolation semantics are that `BEGIN` and `END` are only necessary to ensure the atomicity of _multiple_ statements, but that any single statement which is atomic by itself should not behave differently within or outside of an explicit `BEGIN` / `END`.

To avoid having to think about this subtlety, though, we should always put it in a transaction.

README.md
JoshRosen: I think that Amazon FAQ entry is really misleading / incomplete. My understanding of S3 read-after-write consistency is that 

1. it doesn't apply to bucket listing operations (see https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel), and
2. while all regions _provide_ read-after-write consistency, you have to use certain endpoints to receive those guarantees (specifically, I don't think that the global endpoint provides this for us-standard buckets (I'd clarify this with Amazon if you're really curious about the specifics)).

I don't know why the Amazon documentation doesn't just spell this out more clearly.

JoshRosen:
@yhuai, I've updated this to be organized slightly differently. PTAL.

yhuai:
LGTM

JoshRosen:
Alright, I'm going to merge this now and will create issues to work on the future improvements promised in the text.

PR #151, opened 2016-01-12
Use MANIFEST in unload to guard against eventually-consistent S3 bucket listing

JoshRosen:
This patch modifies `spark-redshift`'s read path in order to guard against a potential source of missing data / inconsistent reads.

According to the [Amazon S3 Data Consistency Model](https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel) documentation, S3 bucket listing is eventually-consistent. In `spark-redshift` 1.6.0 and earlier, the S3 read path performs bucket-listing and thus may be impacted by this eventual-consistency, meaning that in rare circumstances reads may see only a subset of the unloaded data.

This patch fixes this issue by using [`UNLOAD`](https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) command's the `MANIFEST` parameter, which causes it to write a JSON manifest which lists the paths of all unload output files. I modified the UNLOAD query to produce this manifest and added some code to the reader to consume it. In order to simplify parsing of the JSON manifest, I used the [minimal-json](https://github.com/ralfstx/minimal-json) JSON library, chosen because it's small and has no external dependencies (and thus is unlikely to cause the sorts of dependency conflicts that something like Jackson would lead to).

### DIFF
--- a/src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
+++ b/src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
@@ -16,12 +16,13 @@
 
 package com.databricks.spark.redshift
 
-import java.io.File
+import java.io.{ByteArrayInputStream, File}
 import java.net.URI
 
 import com.amazonaws.services.s3.AmazonS3Client
-import com.amazonaws.services.s3.model.BucketLifecycleConfiguration
+import com.amazonaws.services.s3.model.{S3ObjectInputStream, S3Object, BucketLifecycleConfiguration}
 import com.amazonaws.services.s3.model.BucketLifecycleConfiguration.Rule
+import org.apache.http.client.methods.HttpRequestBase
 import org.mockito.Matchers._
 import org.mockito.Mockito
 import org.mockito.Mockito.when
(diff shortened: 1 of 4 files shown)


src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -115,6 +118,26 @@ class RedshiftSourceSuite
       new BucketLifecycleConfiguration().withRules(
         new Rule().withPrefix("").withStatus(BucketLifecycleConfiguration.ENABLED)
       ))
+    val mockManifest = Mockito.mock(classOf[S3Object], Mockito.RETURNS_SMART_NULLS)
```
JoshRosen: This messy mocking is only to get the unit tests to pass. It's not the most high-fidelity mocking in the world, but that's okay because this code path is well-exercised by the integration tests.

JoshRosen:
Ping @liancheng, could you help me to review this?

liancheng:
LGTM

JoshRosen:
Thanks for reviewing; I've updated the README to reflect the new semantics and have fixed the merge conflicts, so I'll merge this as soon as tests pass.

PR #156, opened 2016-01-15
Quote dates and timestamps constants when pushing down filters to Redshift

### LINKED ISSUE #152: DATE/TIMESTAMP filter expression issues
spark 1.6.0, spark-redshift 0.6.0

A few different issues here, seemingly the same place in the code, so lumping them together.
https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala#L44

**DATEs not quoted**
For columns of (Redshift) DATE type, when using the following syntax
`t.filter("c_date = cast('2016-01-01' as DATE)")`
or
`t.filter("c_date = to_date('2016-01-01')")`
or
`t.filter("c_date < current_date()")`
the filter is pushed as part of the UNLOAD command, but the date string ends up not being quoted on the SQL side as such
`WHERE "c_date" = 2016-01-01`
which really results in the following
`WHERE "c_date" = 2014`
because `2016-01-01` is interpreted as a numeric expression (2016 minus 1 minus 1), not a date string.
```
redshift=# select 2016-01-01;
 ?column?
----------
     2014
(1 row)
```
Thus, the wrong results are returned to the user (bad).

**DATE strings not pushed**
Interestingly enough using
`t.filter($"c_date"     >= "2016-01-01")`
does not push any SQL filter through to the UNLOAD command which, while returns the correct result, scans much more data on the Redshift side than need be and should also be fixed.

**TIMESTAMPs not quoted**
If the column type is TIMESTAMP, we have the same problem as DATE when a string is passed in
`t.filter($"c_timestamp" === "2016-01-01 00:00:00")`
results in the following SQL
`WHERE "c_timestamp" = 2016-01-01 00:00:00.0`
which simply throws a Redshift SQL parse error.
Also happens when a Spark TimestampType is passed
`t.filter($"c_timestamp" < current_timestamp())`
which converts to
`WHERE "c_timestamp" < 2016-01-14 15:03:43.928`

It seems the right thing to do here for DATE/TIMESTAMP is check for both types and use the correct casting functions on the database side with the corresponding string format masks.

JoshRosen:
This patch fixes an issue which could lead to incorrect results for queries which contained filters with date or timestamp constants. We need to quote the date and timestamp literals in order to prevent them from being interpreted as arithmetic expressions or producing syntax errors. For more details, see #152.

Fixes #152.

/cc @gregrahn, who reported this issue, and @marmbrus for review.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala
+++ b/src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala
@@ -17,7 +17,7 @@
 package com.databricks.spark.redshift
 
 import org.apache.spark.sql.sources._
-import org.apache.spark.sql.types.{DataType, StringType, StructType}
+import org.apache.spark.sql.types._
 
 /**
  * Helper methods for pushing filters into Redshift queries.
@@ -46,6 +46,7 @@ private[redshift] object FilterPushdown {
      getTypeForAttribute(schema, attr).map { dataType =>
        val sqlEscapedValue: String = dataType match {
          case StringType => s"\\'${value.toString.replace("'", "\\'\\'")}\\'"
+         case DateType | TimestampType => s"\\'$value\\'"
          case _ => value.toString
        }
        s""""$attr" $comparisonOp $sqlEscapedValue"""
--- a/src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala
+++ b/src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala
@@ -52,7 +52,7 @@ private[redshift] object FilterPushdown {
           // Workaround for SPARK-10195: prior to Spark 1.5.0, the Data Sources API exposed internal
           // types, so we must perform conversions if running on older versions:
           if (SPARK_VERSION < "1.5.0") {
-            CatalystTypeConverters.convertToScala(internalValue, DateType)
+            CatalystTypeConverters.convertToScala(internalValue, dataType)
           } else {
             internalValue
           }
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/redshift/FilterPushdown.scala
```diff
@@ -46,6 +46,7 @@ private[redshift] object FilterPushdown {
      getTypeForAttribute(schema, attr).map { dataType =>
        val sqlEscapedValue: String = dataType match {
          case StringType => s"\\'${value.toString.replace("'", "\\'\\'")}\\'"
+         case DateType | TimestampType => s"\\'$value\\'"
```
JoshRosen: I think this might actually be subtly wrong for two reasons:

1. the internal representation of these types might not be Java date or timestamp types; rather, it might be some type of numeric primitive, in which case this would return the wrong result.
2. Java dates and timestamps might have locale-sensitive toString() methods; we should explicitly use a formatter when converting them to strings.

JoshRosen: In Spark 1.5.0+ this will only expose the public SQL types, but in Spark 1.4.x this will return a long for a date; see https://issues.apache.org/jira/browse/SPARK-10195

Here's a test failure which illustrates the problem: https://travis-ci.org/databricks/spark-redshift/jobs/102498199

Therefore, we need to be prepared to handle internal representations here when running on Spark 1.4.x.

JoshRosen: In 1.5+, we'll return java.sql.Date and java.sql.Timestamp, which have locale-independent toString methods, so I think I just have to change this to handle the 1.4.x case and then I think we should be good.

JoshRosen:
Given @gregrahm's LGTM over at #152 plus my own self-review, I'd say that this is good-to-go, so I'm going to merge this now.

Issue #154, opened 2016-01-14
Create temporary table in transaction when performing overwrites


JoshRosen:
Followup to #150:

> **Overwriting an existing table**: By default, `spark-redshift` uses transactions to perform overwrites. Outside of a transaction, it will create an empty temporary table and append the new rows using a `COPY` statement. If the `COPY` succeeds, it will use a transaction to atomically delete the overwritten table and rename the temporary table to destination table.
 >
 > In a future release, this will be changed so that the temporary table is created in the same transaction as the `COPY`.

This can probably be done as part of the same PR which fixes #153

JoshRosen:
Subsumed by #157.

PR #157, opened 2016-01-17
Simplify use of transactions when performing overwrites and when creating new tables

### LINKED ISSUE #153: Issue CREATE TABLE and COPY commands as part of the same transaction when creating a new table
Followup to #150:

> **Creating a new table**: Creating a new table is a two-step process, consisting of a `CREATE TABLE` command followed by a [`COPY`](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) command to append the initial set of rows. Currently, these two steps are performed in separate transactions, so their effects may become visible at different times to readers. The `COPY` itself is atomic, so the table will never be visible in a state where it contains a non-empty subset of the saved rows. In a future release, this will be changed so that the `CREATE TABLE` and `COPY` statements are issued as part of the same transaction.

JoshRosen:
This patch aims to significantly simplify our use of transactions when performing overwrites and when creating new tables; it addresses a few shortcomings in the existing code: previously, we didn't issue the `CREATE TABLE` and `COPY` commands as part of the same transaction when creating new tables and when performing overwrites we created a temporary table outside of a transaction, opening the possibility for that table to be leaked in rare corner-cases.

To address this, this patch refactors the main write logic so that it uses a JDBC connection with auto-commit disabled. Explicitly issuing `commit()` lets us significantly simplify the code and reduces duplication between the branches for handling different SaveModes.

One important change: previously, `usestagingtable` would literally create a separate staging table, but now we will simply delete the existing table and re-create it in a transaction; this frees us of the burden of having to find a unique name for the staging table. This should be safe due to Redshift's serializable snapshot isolation semantics. When `usestagingtable` is disabled, this now means that we simply issue an extra `commit()` after the original table's deletion so that the original table's resources are freed (which lets us write the new rows without having to maintain both the old and new tables at the same time).

/cc @jaley, who wrote the original staging-table-handling code; it would be good if you could confirm whether the changes here will undermine the performance benefits / semantics of the old `usestagingtable=false` path.

Fixes #153 and #154.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -328,6 +328,12 @@ private[redshift] class RedshiftWriter(
         "For save operations you must specify a Redshift table name with the 'dbtable' parameter")
     }
 
+    if (!params.useStagingTable) {
+      log.warn("Setting useStagingTable=false is deprecated; instead, we recommend that you " +
+        "drop the target table yourself. For more details on this deprecation, see" +
+        "https://github.com/databricks/spark-redshift/pull/157")
+    }
+
     val creds: AWSCredentials =
       AWSCredentialsUtils.load(params, sqlContext.sparkContext.hadoopConfiguration)
 
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -47,20 +47,15 @@ import org.apache.spark.sql.types._
  *     non-empty. After the write operation completes, we use this to construct a list of non-empty
  *     Avro partition files.
  *
- *   - Using JDBC, start a new tra
- *
- *   - Use JDBC to issue any CREATE TABLE commands, if required.
- *
  *   - If there is data to be written (i.e. not all partitions were empty), then use the list of
  *     non-empty Avro files to construct a JSON manifest file to tell Redshift to load those files.
  *     This manifest is written to S3 alongside the Avro files themselves. We need to use an
  *     explicit manifest, as opposed to simply passing the name of the directory containing the
  *     Avro files, in order to work around a bug related to parsing of empty Avro files (see #96).
  *
- *   - Use JDBC to issue a COPY command in order to instruct Redshift to load the Avro data into
- *     the appropriate table. If the Overwrite SaveMode is being used, then by default the data
- *     will be loaded into a temporary staging table, which later will atomically replace the
- *     original table via a transaction.
+ *   - Start a new JDBC transaction and disable auto-commit. Depending on the SaveMode, issue
+ *     DELETE TABLE or CREATE TABLE commands, then use the COPY command to instruct Redshift to load
+ *     the Avro data into the appropriate table.
  */
 private[redshift] class RedshiftWriter(
     jdbcWrapper: JDBCWrapper,
@@ -126,6 +121,7 @@ private[redshift] class RedshiftWriter(
         case e: SQLException =>
           // Try to query Redshift's STL_LOAD_ERRORS table to figure out why the load failed.
           // See http://docs.aws.amazon.com/redshift/latest/dg/r_STL_LOAD_ERRORS.html for details.
+          conn.rollback()
           val errorLookupQuery =
             """
               | SELECT *
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
@@ -141,6 +141,8 @@ private[redshift] class RedshiftWriter(
         jdbcWrapper.executeInterruptibly(conn.prepareStatement(copyStatement))
       } catch {
         case e: SQLException =>
+          log.error("SQLException thrown while running COPY query; will attempt to retrieve " +
+            "more information by querying the STL_LOAD_ERRORS table", e)
           // Try to query Redshift's STL_LOAD_ERRORS table to figure out why the load failed.
           // See http://docs.aws.amazon.com/redshift/latest/dg/r_STL_LOAD_ERRORS.html for details.
           conn.rollback()
@@ -356,6 +358,7 @@ private[redshift] class RedshiftWriter(
     } catch {
       case NonFatal(e) =>
         try {
+          log.error("Exception thrown during Redshift load; will roll back transaction", e)
           conn.rollback()
         } catch {
           case NonFatal(e2) =>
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -48,6 +47,8 @@ import org.apache.spark.sql.types._
  *     non-empty. After the write operation completes, we use this to construct a list of non-empty
  *     Avro partition files.
  *
+ *   - Using JDBC, start a new tra
```
JoshRosen: TODO: this entire block comment needs to be updated.

JoshRosen:
I'd love feedback on whether the tests here are adequate; my gut says that we need a test which counts the number of times that `commit()` is called when not using a staging table. I can add this tomorrow.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -360,19 +316,35 @@ private[redshift] class RedshiftWriter(
...
+        if (!params.useStagingTable) {
+          // If we're not using a staging table, commit now so that Redshift doesn't have to
+          // maintain a snapshot of the old table during the COPY; this sacrifices atomicity for
+          // performance.
+          conn.commit()
+        }
       }
+      log.info(s"Loading new Redshift data to: $table")
+      doRedshiftLoad(conn, data, params, creds, manifestUrl)
+      conn.commit()
```
JoshRosen: Does this need to handle InterruptedException? i.e. can this block for a really long time?

JoshRosen:
Ping @jaley, do you have any cycles to help review this? I think that you might be a good person to look at this because you have a lot of the context RE: the original implementation of this functionality.

jaley:
Hey @JoshRosen ,

Sure. Sorry for being slow to notice this - been playing catch up after a long vacation. I'll take a look in the next day or so if that's cool with you?

JoshRosen:
Sure, that's fine.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
jaley: It has been my experience that virtually any Redshift command, `COMMIT` included, can block for several minutes under certain circumstances. I think the most likely cause is that the cluster has WLM parameters configured to put the connected client on a limited pool of some type, such that all commands will be queued when all slots are taken by other queries.

I have a feeling that means though, if you plan to send a `ROLLBACK` command in response to the interruption, that will also block for many minutes...

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
```diff
@@ -360,19 +312,35 @@ private[redshift] class RedshiftWriter(
...
+    } catch {
+      case NonFatal(e) =>
+        try {
+          conn.rollback()
```
jaley: Might a log.info or log.warn be good here, to inform anyone testing their application code that the load failed and what they're now waiting for is a rollback to finish?

jaley:
Do we currently run the integration test with just the Redshift JDBC driver? Now that we're using the driver to manage transactions, I wonder if we might start to see different behaviour between the Redshift and Postgres drivers? Amazon's recommendation is to use their official driver, but I think we've previously been lucky and users could safely make use of either for the small set of commands we rely on. This change effectively exercises different code paths in the driver, so it might be worth making sure that if anything no-longer works about the Postgres driver, we're at least able to communicate that clearly to users and explicitly state that it's not supported any more. (Hopefully it's still fine anyway!)

jaley:
The change looks good to me. I wonder if perhaps the `useStagingTable` parameter name is now a little confusing? This is certainly the equivalent behaviour, and so in the interest of maintaining compatibility it should certainly be supported, but I suspect if we were choosing a name for that now, we probably wouldn't refer to staging tables.

I also wonder if it really communicates the fact that if you set this to false, you save disk space but you run the risk of losing your existing data? If not, that was always the case, this change just caused me to think about it some more!

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -325,23 +318,21 @@ class RedshiftSourceSuite
...
     )
 
     val source = new DefaultSource(mockRedshift.jdbcWrapper, _ => mockS3Client)
     intercept[Exception] {
       source.createRelation(testSqlContext, SaveMode.Overwrite, params, expectedDataDF)
-      mockRedshift.verifyThatConnectionsWereClosed()
-      mockRedshift.verifyThatExpectedQueriesWereIssued(expectedCommands)
     }
     mockRedshift.verifyThatConnectionsWereClosed()
+    mockRedshift.verifyThatCommitWasNotCalled()
```
emlyn: Should this also verify that rollback was called?

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
JoshRosen: I'm not sure that it's safe for us to wrap this in order to catch InterruptedException since I don't think that it's safe to call `.rollback()` while the other thread is in the middle of executing `commit()`. Therefore, I'm going to leave this as-is for now and will revisit later if this turns out to be a problem in practice.

src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
JoshRosen: Done.

src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala
JoshRosen: Done.

JoshRosen:
@jaley:

> The change looks good to me. I wonder if perhaps the useStagingTable parameter name is now a little confusing? This is certainly the equivalent behaviour, and so in the interest of maintaining compatibility it should certainly be supported, but I suspect if we were choosing a name for that now, we probably wouldn't refer to staging tables.

I think that we should just deprecate this parameter. If a user wants to avoid the performance penalty of using a transaction for overwrites then I don't think it's unreasonable to suggest that they drop the staging table themselves.

> I also wonder if it really communicates the fact that if you set this to false, you save disk space but you run the risk of losing your existing data? If not, that was always the case, this change just caused me to think about it some more!

Yep, this is the old behavior. It looks like you documented this in one of the original commits that added this functionality: https://github.com/databricks/spark-redshift/blame/v0.5.0/README.md#L280

JoshRosen:
(Err, that should be drop the _old_ table, not the staging table)

JoshRosen:
I'm going to go ahead and merge this now and will address any minor comments in folllowups. I'll give this a test with both drivers and then will begin work on packaging a new release.

PR #159, opened 2016-01-18
Have AWSCredentialsUtils.load() perform the check for STS credentials

JoshRosen:
This patch moves the check for STS credentials into the `AWSCredentials.load()` call itself. The motivation for this is a single call-site where we forgot to do a `.getOrElse()`, which caused the STS credentials to be ignored. This broke `spark-redshift` for users who authenticated to AWS only via IAM instance profiles.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -20,7 +20,6 @@ import java.net.URI
 
 import com.amazonaws.auth.AWSCredentials
 import com.amazonaws.services.s3.AmazonS3Client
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -85,8 +84,7 @@ private[redshift] case class RedshiftRelation(
   }
 
   override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
-    val creds =
```
JoshRosen: This is the callsite which was missing the `params.temporaryAWSCredentials.getOrElse()` call.

PR #163, opened 2016-01-25
Update artifact coordinates in README to include _2.11 and _2.10 suffixes

julienbaley:
The current coordinates do not point to an existing package. Copying from https://github.com/databricks/spark-csv/blob/master/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,17 @@ This library is more suited to ETL than interactive queries, since large amounts
 
 You may use this library in your applications with the following dependency information:
 
+**Scala 2.10**
 ```
 groupId: com.databricks
-artifactId: spark-redshift
+artifactId: spark-redshift_2.10
+version: 0.6.0
+```
+
+**Scala 2.11**
+```
+groupId: com.databricks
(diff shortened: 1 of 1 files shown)


JoshRosen:
1.3.0 is not an existing version; see http://mvnrepository.com/artifact/com.databricks/spark-redshift_2.10 and http://mvnrepository.com/artifact/com.databricks/spark-redshift_2.11

julienbaley:
Woops, sorry, updated that :)

README.md
```diff
@@ -26,9 +26,16 @@ This library is more suited to ETL than interactive queries, since large amounts
 
 You may use this library in your applications with the following dependency information:
 
+### Scala 2.10
```
JoshRosen: Super-minor nit: mind changing this to be bold text (with `**`) instead of a subheading so that it doesn't mess up my table-of-contents generation?

julienbaley: Good point. Done

JoshRosen:
Thanks for updating. LGTM, so I'm going to merge this.

PR #165, opened 2016-01-28
Add CSV option

emlyn:
We've got to the stage where our jobs are big enough that Redshift loads are becoming a major bottleneck with its AVRO performance problems. We're hoping Amazon will fix them, but in the mean time we've started running with a modified spark-redshift to allow us to import via CSV (Redshift `COPY`s are running at least 5 times faster with this for us).
It's probably not a good idea to merge this in at the moment - I'm not sure how robust it is to different data types / unusual characters in strings etc., and it would at least need some tests and documentation. And hopefully Amazon will soon fix AVRO import making it unnecessary anyway. But I thought I'd share my changes in case they are useful to anyone else.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Parameters.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Parameters.scala
@@ -31,11 +31,15 @@ private[redshift] object Parameters {
     // * distkey has no default, but is optional unless using diststyle KEY
     // * jdbcdriver has no default, but is optional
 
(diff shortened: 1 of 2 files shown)


JoshRosen:
Thanks for sharing. I'm glad to see that this was a relatively small change.

I agree that it's probably best to wait and see if Amazon speeds up Avro ingest; let's wait a couple of months and re-assess this feature later if there's significant interest / demand.

PR #168, opened 2016-02-01
Updated error message for lifecycle case None

mamdouhweb:
Updated error message for lifecycle case None and changed the log level to warning

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Utils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Utils.scala
@@ -94,8 +94,8 @@ private[redshift] object Utils {
       Option(s3Client.getBucketLifecycleConfiguration(bucket)) match {
         case None =>
           log.warn(s"No configuration were found for S3 bucket $bucket. " +
-             "Please make sure that the bucket exists and it has a Life Cycle " +
-             "configuration applied")
+             "Please make sure that the bucket exists and it has a lifecycle " +
+             "configuration applied.")
         case Some(lifecycleConfiguration) =>
           val someRuleMatchesTempDir = lifecycleConfiguration.getRules.asScala.exists { rule =>
             // Note: this only checks that there is an active rule which matches the temp directory;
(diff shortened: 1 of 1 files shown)


JoshRosen:
This is a decent start, but I'd do this a little differently. I have a better warning message written a few lines down; do you think you could refactor the conditions here so that both warning cases hit the same existing warning method instead?

mamdouhweb:
Agree, the point though is that None indicates either not having configuration at all or that the bucket does not exists where Some tells you that we have found some configuration and LifeCycle config might be one of them, then the check you do elaborates on whether it's been setup correctly or not.

For the sake of not having one case looking into both responsibilities, I would say let's keep them separate, where the message I proposed calls it explicitly that No Configuration were found at all, and then recommends setting up the LifeCycle ones in case the bucket exists, and ask to verify the existence of the bucket in case it wasn't created.

Thoughts?

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -93,7 +93,9 @@ private[redshift] object Utils {
       val key = Option(s3URI.getKey).getOrElse("")
       Option(s3Client.getBucketLifecycleConfiguration(bucket)) match {
         case None =>
-          log.error(s"The S3 bucket $bucket does not exist")
+          log.warn(s"No configuration were found for S3 bucket $bucket. " +
+             "Please make sure that the bucket exists and it has a Life Cycle " +
```
JoshRosen: Can you please try to match the writing style of the existing messages? "Life Cycle" should be "lifecycle" and this sentence should end with a period.

mamdouhweb: done. :)

JoshRosen:
The new error messages still has some annoying properties; "no configuration were found for S3 bucket" is still ambiguous / confusing. We're not paying per-word, so it's fine to be a bit more specific / verbose in the warning message. It would also be nice if the error message contained the link to the AWS docs.

Also, if we _are_ going to distinguish between the error messages in these two branches then the `Some` case should explain that a lifecycle configuration was found but that no rule matched the `tempdir` path. However, I'm not sure that this distinction actually matters and therefore would prefer to just combine it into a single message. I'll do this myself in a separate PR.

JoshRosen:
I also don't think that we need to explicitly say that the bucket doesn't exist since that case will be handled downstream.

JoshRosen:
I've opened #177 to replace this.

PR #179, opened 2016-02-12
Update README to clearly separate the library name from its maven coordinates

mengxr:
Calling the library `spark-redshift` might confuse users about the source of the library, which doesn't conform to ASF trademark policy. This PR names the library `Redshift data source for Spark` and removes `spark-redshift` references from README.


JoshRosen:
Just realized that we probably also have to update `tutorial/README`.

PR #196, opened 2016-04-01
Log effective Redshift COPY statement

jacek-rzrz:
This seems like a useful log message because:

 - It's reassuring to see that params set via extracopyoptions are
   actually used.

 - Currently the last log message one can see while COPY executes is
   CREATE TABLE logged a few lines before. This is a little confusing as
   can be understood as CREATE TABLE statement taking a long time.

 - There is a symmetric log message for UNLOAD statement in
   RedshiftRelation.scala.


JoshRosen:
LGTM, so I'm going to merge this now. Sorry for the delay.

PR #204, opened 2016-04-18
Fix decimal precision loss when reading the results of a Redshift query

### LINKED ISSUE #203: precision losing while changing data in select with limit
my data looks as follow

```
+----+--------+---------+
|  id|latitude|longitude|
+----+--------+---------+
|8919|15314215| 91593373|
+----+--------+---------+
```

i ran query as 

``` python
query = "select id, latitude, longitude from table"
>>> df = sql_context.read \
...    .format("com.databricks.spark.redshift") \
...    .option("url", redshift_jdbc_url) \
...    .option("query", query) \
...    .option("tempdir", AWS_S3_BUCKET) \
...    .load()
>>> df.show(1)
+----+--------+---------+
|  id|latitude|longitude|
+----+--------+---------+
|8919|15314215| 91593373|
+----+--------+---------+
>>> df.printSchema()
root
 |-- id: long (nullable = true)
 |-- latitude: long (nullable = true)
 |-- longitude: long (nullable = true)
```

but when i put `latitude/1000000.0 as latitude` its breaking

``` python

query = "select id, latitude/1000000.0 as latitude, longitude/1000000.0 as longitude from table"
>>> df = sql_context.read \
...    .format("com.databricks.spark.redshift") \
...    .option("url", redshift_jdbc_url) \
...    .option("query", query) \
...    .option("tempdir", AWS_S3_BUCKET) \
...    .load()
>>> df.show(1)
+----+--------+---------+
|  id|latitude|longitude|
+----+--------+---------+
|8919|      15|       91|
+----+--------+---------+
>>> df.printSchema()
root
 |-- id: long (nullable = true)
 |-- latitude: decimal(18,0) (nullable = true)
 |-- longitude: decimal(18,0) (nullable = true)
```

JoshRosen:
This patch addresses a somewhat obscure corner-case bug where we would lose decimal precision when using `query` to load the results of a Redshift query. For an example of this, see #203.

The root problem is that the `SELECT * FROM $table WHERE 1=0` query that we were using to query Redshift schemas seems to affect the precision of decimal columns. If I replace the `WHERE 1=0` with a `LIMIT 1` (since `LIMIT 0` has the same problem), then Redshift reports the correct precision.

Fixes #203.


src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -165,7 +165,11 @@ private[redshift] class JDBCWrapper {
    * @throws SQLException if the table contains an unsupported type.
    */
   def resolveTable(conn: Connection, table: String): StructType = {
-    val rs = executeQueryInterruptibly(conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0"))
+    // We need the `LIMIT 1` here to work around a corner-case bug where a query which returns no
+    // rows may sometimes cause Redshift to report the wrong decimal precision, leading to a loss of
+    // precision when we apply this schema to the results of the full unload. For more details,
+    // see https://github.com/databricks/spark-redshift/issues/203
+    val rs = executeQueryInterruptibly(conn.prepareStatement(s"SELECT * FROM $table LIMIT 1"))
```
JoshRosen: This might have undesirable consequences if the table contains individual rows with gigantic values, such as a 100 megabyte blob column. I wonder if there's a way to avoid that...

JoshRosen:
Alright, going to merge this now that the changes in #301 have reduced the potential for perf. regression here.

Issue #199, opened 2016-04-06
Support for IAM roles with Redshift's COPY command


ravgit:
Redshift recently added support for IAM roles with the COPY and UNLOAD commands (pls see below). 
https://aws.amazon.com/releasenotes/Amazon-Redshift/8869404215180370
http://docs.aws.amazon.com/redshift/latest/mgmt/authorizing-redshift-service.html

This makes the COPY, UNLOAD operations relatively more secure by simply having to specify 'aws_iam_role' parameter with the COPY command instead of the 'access_key' and 'secret_access_key'. I would like to request that an 'aws_iam_role' option be added to the spark-redshift package so we can use the roles instead of the access/secret keys. Here is an example of what I would like to do.

scala> val iamrole = "arn:aws:iam::123456789000:role/RedshiftCopyUnload"
iamrole: String = arn:aws:iam::123456789000:role/RedshiftCopyUnload

scala> val testDF = sqlContext.read.format("com.databricks.spark.redshift").option("url", jdbcURL).option("tempdir", tempS3Dir).option("dbtable", "flights").option("aws_iam_role", iamrole).load()

JoshRosen:
This looks pretty straightforward. I'd probably go with a design where a provided IAM role takes precedence over any other authentication option.

Want to take a shot at implementing this and submit a PR? I imagine that this will likely replace / subsume some of the existing IAM instance profile README instructions, so if it was me I'd start by trying to write the README update as an exercise to see whether the new API will introduce any ambiguity or break compatibility in any way.

blrnw3:
This is a lot nicer than generating temp credentials every time 👍

PR #216, opened 2016-06-16
 Support for IAM roles with Redshift's COPY command

smallcrew:
This patch allows us to copy from redshift using iam roles
see #199


blrnw3:
:+1 @databricks

JoshRosen:
Thanks for the patch. I'm going to close this in favor of #219, which addresses the same issue but has more comprehensive documentation. Also, it looks like there are multiple sets of unrelated changes that became mixed together in this PR; try re-opening your other PR from a unique topic branch rather than from the `master` branch of your fork. Thanks!

PR #218, opened 2016-06-22
Specify extra unloadoptions

smallcrew:
Allow user to specify unload options like a custom delimiter for example
By default, "manifest" and "escape" options are always set

Example:
`.option("extraunloadoptions", "DELIMITER ','")`

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,19 @@ for other options).</p>
 at the end of the command can be used, but that should cover most possible use cases.</p>
     </td>
  </tr>
+ <tr>
+    <td><tt>extraunloadoptions</tt></td>
+    <td>No</td>
+    <td>No Default</td>
+    <td>
+<p>A list extra unload options to append to the Redshift <tt>UNLOAD</tt> command when writing data to S3, e.g. <tt>DELIMITER</tt>
+or <TT>FIXEDWIDTH</tt> (see the <a href="http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html#unload-parameters">Redshift docs</a>
+for other options). Parameters <tt>ESCAPE</tt> and <tt>MANIFEST</tt> are always append.</p>
+
+<p>Note that since these options are appended to the end of the <tt>UNLOAD</tt> command, only options that make sense
+at the end of the command can be used, but that should cover most possible use cases.</p>
+    </td>
+ </tr>
 </table>
 
 ## Additional configuration options
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -179,7 +179,12 @@ private[redshift] case class RedshiftRelation(
     // the credentials passed via `credsString`.
     val fixedUrl = Utils.fixS3Url(Utils.removeCredentialsFromURI(new URI(tempDir)).toString)
 
-    s"UNLOAD ('$query') TO '$fixedUrl' WITH CREDENTIALS '$credsString' ESCAPE MANIFEST"
+    val userUnloadOptions = params.parameters.getOrElse("extraunloadoptions", "")
+    val unloadOptions = (userUnloadOptions ++ Array("ESCAPE", "MANIFEST")
+      .filterNot(r => userUnloadOptions.contains(r))
+      .mkString(" ", " ", "")).trim
+
+    s"UNLOAD ('$query') TO '$fixedUrl' WITH CREDENTIALS '$credsString' $unloadOptions"
   }
 
   private def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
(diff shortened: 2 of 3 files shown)


JoshRosen:
This seems reasonable, but I have a few suggestions:

- [ ] Name this `extraunloadoptions` for symmetry with the existing `extracopyoptions`.
- [ ] Document it in the README.
- [ ] Remove the unrelated whitespace changes.

JoshRosen:
Friendly ping. @smallcrew, will you have time to finish this PR and incorporate my feedback?

smallcrew:
Hi @JoshRosen feedback incorporated

README.md
```diff
@@ -414,6 +414,19 @@ for other options).</p>
 at the end of the command can be used, but that should cover most possible use cases.</p>
     </td>
  </tr>
+ <tr>
+    <td><tt>extraunloadoptions</tt></td>
+    <td>No</td>
+    <td><tt>ESCAPE</tt> <tt>MANIFEST</tt> </td>
```
JoshRosen: Minor subtlety: I wouldn't necessarily say that these are defaults for this setting because if the user sets `extraunloadoptions` to some other value it doesn't override / remove `ESCAPE` and `MANIFEST`. Note that `extracopyoptions`' doc says "No Default" so I wonder if we should do the same here in order to avoid confusion.

JoshRosen: Also, `spark-redshift` won't work unless `MANIFEST` is set, so I don't want to let users remove it.

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -179,7 +179,12 @@ private[redshift] case class RedshiftRelation(
     // the credentials passed via `credsString`.
     val fixedUrl = Utils.fixS3Url(Utils.removeCredentialsFromURI(new URI(tempDir)).toString)
 
-    s"UNLOAD ('$query') TO '$fixedUrl' WITH CREDENTIALS '$credsString' ESCAPE MANIFEST"
+    val userUnloadOptions = params.parameters.getOrElse("extraunloadoptions", "").toUpperCase()
```
JoshRosen: I'm not sure that you necessarily want to automatically convert this to upper-case; do you remember why this was necessary?

JoshRosen: For example, what if you wanted the delimiter to be a lowercase-character?

smallcrew:
@JoshRosen You are right, changes done

src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
```diff
@@ -179,7 +179,12 @@ private[redshift] case class RedshiftRelation(
     // the credentials passed via `credsString`.
     val fixedUrl = Utils.fixS3Url(Utils.removeCredentialsFromURI(new URI(tempDir)).toString)
 
-    s"UNLOAD ('$query') TO '$fixedUrl' WITH CREDENTIALS '$credsString' ESCAPE MANIFEST"
+    val userUnloadOptions = params.parameters.getOrElse("extraunloadoptions", "")
+    val unloadOptions = (userUnloadOptions ++ Array("ESCAPE", "MANIFEST")
+      .filterNot(r => userUnloadOptions.contains(r))
```
JoshRosen: What's this filter doing? This is a bit confusing.

README.md
```diff
@@ -414,6 +414,19 @@ for other options).</p>
...
     </td>
  </tr>
+ <tr>
+    <td><tt>extraunloadoptions</tt></td>
+    <td>No</td>
+    <td>No Default</td>
+    <td>
+<p>A list extra unload options to append to the Redshift <tt>UNLOAD</tt> command when writing data to S3, e.g. <tt>DELIMITER</tt>
+or <TT>FIXEDWIDTH</tt> (see the <a href="http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html#unload-parameters">Redshift docs</a>
+for other options). Parameters <tt>ESCAPE</tt> and <tt>MANIFEST</tt> are always append.</p>
```
JoshRosen: I'd say

> The <tt>ESCAPE</tt> and <tt>MANIFEST</tt> are options always automatically appended to the <tt>UNLOAD</tt> query and thus should not be specified here.</p>

README.md
```diff
@@ -414,6 +414,19 @@ for other options).</p>
 at the end of the command can be used, but that should cover most possible use cases.</p>
     </td>
  </tr>
+ <tr>
+    <td><tt>extraunloadoptions</tt></td>
+    <td>No</td>
+    <td>No Default</td>
+    <td>
+<p>A list extra unload options to append to the Redshift <tt>UNLOAD</tt> command when writing data to S3, e.g. <tt>DELIMITER</tt>
```
JoshRosen: > A list _of_ extra ...

PR #219, opened 2016-06-24
Add IAM role support for COPY/UNLOAD operations

### LINKED ISSUE #199: Support for IAM roles with Redshift's COPY command

jonbender:
This fixes #199 

I have tested it out with one of our Redshift clusters and it seems to behave as expected.  Also I realized too late that there was a similar patch up (https://github.com/databricks/spark-redshift/pull/216/files).  Please feel free to close if you prefer that one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,8 +196,7 @@ This library reads and writes data to S3 when transferring data to/from Redshift
 
 There are four ways of configuring AWS credentials for use by this library:
 
-1. **Set the `aws_iam_role` parameter:** If set, this takes precedence over any other authentication option.  You will need to have this IAM role attached to the Redshift cluster which allows read/write access to your `tempdir` bucket.  More info [here](http://docs.aws.amazon.com/redshift/latest/mgmt/copy-unload-iam-role.html)
-2. **Set keys in Hadoop conf (best option for most users):** You can specify AWS keys via [Hadoop configuration properties](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md). For example, if your `tempdir` configuration points to a `s3n://` filesystem then you can set the `fs.s3n.awsAccessKeyId` and `fs.s3n.awsSecretAccessKey` properties in a Hadoop XML configuration file or call `sc.hadoopConfiguration.set()` to mutate Spark's global Hadoop configuration.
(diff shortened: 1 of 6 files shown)


blrnw3:
👍 @databricks

JoshRosen:
I'm inclined to merge this version of the patch instead of #216's because it has good documentation and doesn't have unrelated changes + style errors mixed in. Let me go ahead and test this out in my environment. It would be cool to add an end-to-end integration test for this, but I'll take care of that myself since it's going to require a bit of modification to our test Redshift instance in order to assign it the appropriate IAM role.

README.md
```diff
@@ -173,9 +194,10 @@ val records = sc.newAPIHadoopFile(
...
 
-There are three ways of configuring AWS credentials for use by this library:
+There are four ways of configuring AWS credentials for use by this library:
 
-1. **Set keys in Hadoop conf (best option for most users):** You can specify AWS keys via [Hadoop configuration properties](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md). For example, if your `tempdir` configuration points to a `s3n://` filesystem then you can set the `fs.s3n.awsAccessKeyId` and `fs.s3n.awsSecretAccessKey` properties in a Hadoop XML configuration file or call `sc.hadoopConfiguration.set()` to mutate Spark's global Hadoop configuration.
+1. **Set the `aws_iam_role` parameter:** If set, this takes precedence over any other authentication option.  You will need to have this IAM role attached to the Redshift cluster which allows read/write access to your `tempdir` bucket.  More info [here](http://docs.aws.amazon.com/redshift/latest/mgmt/copy-unload-iam-role.html)
```
JoshRosen: Nit: I would re-order this so that "Set keys in Hadoop conf (best option for most users)" is still the first item in the list, since I think that's the absolute easiest way to get started. I'd place this as option number three, ahead of the existing IAM instance profile support.

JoshRosen: If you don't get to it I might just fix this up myself on merge as part of some general pre-release README cleanup.

JoshRosen:
LGTM. I tested this out with a new integration test suite at https://github.com/databricks/spark-redshift/commit/315a5018eca09e048f8a1e11a7701e40bdb05525 and it worked perfectly, so I'm going to merge this now and will open my own PR to add that new test suite.

PR #226, opened 2016-07-06
Add end-to-end integration tests for aws_iam_role

JoshRosen:
This patch adds an end-to-end integration test which exercises the `aws_iam_role` authentication method added in #219.


JoshRosen:
Failed test is known to be transiently flaky, so I'm going to merge this despite it.

PR #227, opened 2016-07-07
Fix precision loss when unloading timestamps from Redshift

### LINKED ISSUE #214: Sub-second resolution for timestamps lost on load
I have a redshift table with a timestamp column (t_stamp) containing times at sub second resolution, i.e.

2016-05-27 00:27:41.0
2016-05-27 00:27:41.1
2016-05-27 00:27:41.2
2016-05-27 00:27:41.3

The times are of data type "timestamp" in redshift.

I'm using spark-redshift to create a dataframe from the redshift table using 

val df  = sqlContext.read
    .format("com.databricks.spark.redshift")....

and the datafame is loaded into df. The t_stamp column has type "string" now.

It seems as if the sub-second time information in t_stamp is lost and every time stamp is "floored" to the nearest second, i.e.

2016-05-27 00:27:41.0
2016-05-27 00:27:41.0
2016-05-27 00:27:41.0
2016-05-27 00:27:41.0

Is this expected behavior for spark-redshift? It is important for us to have sub-second time resolution for our application.

JoshRosen:
This patch fixes a bug which led to loss of sub-second precision when unloading certain timestamps from Redshift.

In a nutshell, the problem is that Redshift's unloaded timestamps carry only the significant figures in sub-second components of the timestamp, so a timestamp such as `2016-05-27 00:27:41.100` is output as `2016-05-27 00:27:41.1` in the unload output, while one like `2016-05-27 00:27:41.123` is roundtripped with three significant sub-second digits.

From its earliest versions, `spark-redshift` incorrectly assumed that Redshift's unloaded timestamps would conform to either a `yyyy-MM-dd HH:mm:ss.SSS` or `yyyy-MM-dd HH:mm:ss` format, with either three full digits of sub-second precision or none ([source](https://github.com/databricks/spark-redshift/blob/141cf19901a4ad92e824c4fd337e2321d9b597f7/src/main/scala/com/databricks/spark/redshift/Conversions.scala#L51)), so the length of the timestamp string was used to select the DateFormat used for parsing. As a result, timestamps like `2016-05-27 00:27:41.1` would be routed to the "seconds-only" format, losing the sub-second information and leading to the bug reported in #204.

Note, however, that we can't simply fix this by using a different strategy to select the DateFormat (or a greater number of DateFormats); see https://stackoverflow.com/questions/7159895 for a full explanation of why this is the case.

Instead, this patch addresses this issue by using [`java.sql.Timestamp.valueOf`](https://docs.oracle.com/javase/6/docs/api/java/sql/Timestamp.html#valueOf(java.lang.String)), which properly handles fractional seconds,  to parse timestamps. I also removed a redundant date formatter implementation which was used for `Date` columns and have fixed a number of outdated code comments.

To test my fix, I added both end-to-end regression tests as well as unit tests in `ConversionsSuite`.

Fixes #214.

/cc @jaley @mwho

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -640,4 +640,17 @@ class RedshiftIntegrationSuite extends IntegrationSuiteBase {
     // filter pushdown.
     checkAnswer(df.filter("testtimestamp = '2015-07-01 00:00:00.001'"), Seq(Row(timestamp)))
   }
+
+  test("full timestamp precision is preserved in loads (regression test for #214)") {
+    val timestamps = Seq(
+      TestUtils.toTimestamp(1970, 0, 1, 0, 0, 0, millis = 1),
+      TestUtils.toTimestamp(1970, 0, 1, 0, 0, 0, millis = 10),
+      TestUtils.toTimestamp(1970, 0, 1, 0, 0, 0, millis = 100),
+      TestUtils.toTimestamp(1970, 0, 1, 0, 0, 0, millis = 1000))
+    testRoundtripSaveAndLoad(
+      s"full_timestamp_precision_is_preserved$randomSuffix",
+      sqlContext.createDataFrame(sc.parallelize(timestamps.map(Row(_))),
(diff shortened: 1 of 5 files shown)


JoshRosen:

PR #228, opened 2016-07-07
Escape backslashes in queries passed to UNLOAD

### LINKED ISSUE #215: Backslashes in load queries are not escaped
I get a syntax error when I try to perform a load using a query that contains backslashes as literals, e.g.

```
SELECT replace(mycol, '\\', '') as col FROM mytable
```

This query gets escaped by spark-redshift (within an `UNLOAD` statement) as follows:

```
SELECT replace(mycol, \'\\\', \'\') as col FROM mytable
```

Notice that only the single quotes are escaped, while both the single quotes and the backslashes need to be escaped, as per the redshift [docs](http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html):

```
If your query contains quotes (enclosing literal values, for example),
or backslashes (\), you need to escape them in the query text ...
```

Using the example above, the escaped query should look like:
```
SELECT replace(mycol, \'\\\\\', \'\') as col FROM mytable
```

One workaround is to add extra backslashes myself to the query. Works but definitely not a long term solution.

JoshRosen:
We need to escape backslashes that appear in the query passed to the `UNLOAD` command. This patch adds a regression test plus additional backslash escaping. Fixes #215.

### DIFF
--- a/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
+++ b/src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
@@ -186,6 +186,19 @@ class RedshiftIntegrationSuite extends IntegrationSuiteBase {
     )
   }
 
+  test("backslashes in queries/subqueries are escaped (regression test for #215)") {
+    val loadedDf = sqlContext.read
+      .format("com.databricks.spark.redshift")
+      .option("url", jdbcUrl)
+      .option("query", s"select replace(teststring, '\\\\', '') as col from $test_table")
+      .option("tempdir", tempDir)
+      .load()
+    checkAnswer(
+      loadedDf.filter("col = 'asdf'"),
+      Seq(Row("asdf"))
+    )
+  }
+
   test("Can load output when 'dbtable' is a subquery wrapped in parentheses") {
     // scalastyle:off
     val query =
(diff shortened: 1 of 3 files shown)


src/test/scala/com/databricks/spark/redshift/RedshiftSourceSuite.scala
```diff
@@ -182,10 +182,10 @@ class RedshiftSourceSuite
         |UNLOAD \('SELECT "testbyte", "testbool" FROM
         |  \(select testbyte, testbool
         |    from test_table
-        |    where teststring = \\'Unicode\\'\\'s樂趣\\'\) '\)
+        |    where teststring = \\'\\\\\\\\Unicode\\'\\'s樂趣\\'\) '\)
```
JoshRosen: Note that in this unit test this line is a regex matching expected output, hence the extra level of backslash escaping.

src/it/scala/com/databricks/spark/redshift/RedshiftIntegrationSuite.scala
```diff
@@ -186,6 +186,19 @@ class RedshiftIntegrationSuite extends IntegrationSuiteBase {
     )
   }
 
+  test("backslashes in queries/subqueries are escaped (regression test for #215)") {
```
JoshRosen: This is an end-to-end integration test.

JoshRosen:
Going to merge this fix.

PR #221, opened 2016-06-27
Migrate to Spark 2.0

liancheng:
This PR migrates spark-redshift to Spark 2.0.

It is marked as WIP because it depends on https://github.com/databricks/spark-avro/pull/137.

Issue #197, opened 2016-04-01
Allow asynchronous COPY


jacek-rzrz:
This is probably a long shot but I thought I can ask about it here anyway.

Saving a `DataFrame` to Redshift can be seen as a 2-phase process:

1. Writing files to S3
2. COPYing data from S3 to a Redshift table

Of course the driver program is blocked on `DataFrameWriter.save()` for the duration of both phases.
During phase 2 no Spark jobs are running because things happen outside the cluster, i.e. between S3 and Redshift.
In my setup phase 2 takes much longer than phase 1 and amounts to substantial amount of time.
Perhaps this is uncommon, but I don't actually need to wait for the data to materialize in Redshift to proceed with next jobs in my driver program.

I am wondering if it would be a compelling feature to allow step 2 to happen asynchronously? I would imagine that in such scenario e.g. a `Future` representing phase 2 could somehow be passed to client code however I don't think Data Sources API allows for something like that.

If such a feature would be of interest and if we found a sane way to do it - I would be happy to implement.

JoshRosen:
Today you can already wrap the `DataFrameWriter.save()` call in a Future and it should work as you would expect because Spark and `spark-redshift` should be thread-safe. However, I'm guessing that this might not be exactly what you want in case you only want the second `COPY` phase to be asynchronous.

As you've observed, the Data Sources API doesn't really have a concept that this type of asynchrony would map onto. The only approach that I can think of would be to expose a separate set of lower-level public APIs for the various phases of `spark-redshift` (on top of which we'd have the existing DataFrameWriter / Reader implementations) and define objects / case-classes for holding the data that's passed between these phases (manifest paths, temp directories, etc.). This is probably going to be a big change, so if you're interested in working on something like this I'd spend some time on up-front design to build clean and stable interfaces.

jacek-rzrz:
Hi Josh,
Yes exposing distinct phases sounds good - I would definitely be interested to work on this. Can't devote any time to it at the moment, but starting August I could certainly spend a few hours a week at least.

PR #239, opened 2016-07-17
Cleans up deprecated APIs

### LINKED ISSUE #66: Remove deprecated APIs before 0.7.0 release
Before releasing 0.7.0, we should audit the APIs deprecated in #65 and remove them if nobody has asked us to keep them.

liancheng:
This PR branch is based on databricks/spark-redshift#221 and contains its changes. For easier review, please refer to https://github.com/liancheng/spark-redshift/pull/1.

This PR cleans up all the deprecated APIs in 1.x:

- `Parameters.overwrite`

  Users should use `SaveMode` instead.

- `SchemaParser`

  Removed since the only API method uses this class is removed.

- `RedshiftDataFrame.saveAsRedshiftTable(parameters: Map[String, String])`

  Users should use `DataFrameWriter` methods instead.

- `RedshiftContext.redshiftTable(parameters: Map[String, String])`

  Users should use `DataFrameReader` methods instead.

- `RedshiftContext.redshiftFile(path: String, schema: String)`

  Removed since we'd like to stop using string schema parsing. The following API method is added to replace it:

  ```scala
  def redshiftFile(path: String, schema: StructType): DataFrame
  ```

Fixes #66.


liancheng:
@JoshRosen Since the following public method in `RedshiftContext` isn't marked as deprecated

```scala
def redshiftFile(path: String, columns: Seq[String]): DataFrame
```

I added a new version of `redshiftFile()`

```scala
def redshiftFile(path: String, schema: StructType): DataFrame
```

to replace the deprecated one:

```scala
def redshiftFile(path: String, schema: String): DataFrame
```

These methods are used to read data files unloaded from a Redshift table using `RedshiftInputFormat`, and can't be replaced by existing DataFrame reader/writer API. Another choice can be to make  `RedshiftRelation` support reading unloaded files, so that we can have a more unified API. What do you think?

JoshRosen:
@liancheng, in the long run I think it would make sense to open up more modularized APIs to the `spark-redshift` internals in order to address advanced use-cases (such as the one in #197), so at some point I think we'll want to have a separate component for reading unloaded files. I don't think that we necessarily want to expose that functionality publically through RedshiftRelation since it's not the easiest class to construct AFAIK.

Regarding the individual deprecations:

- +1 on removing `Parameters.overwrite`; this is an obvious good chose.
- +1 on removing `saveAsRedshiftTable` and `redshiftTable`, since these are subsumed by the data sources API.
- +1 on replacing `redshiftFile(path: String, schema: String)` by `redshiftFile(path: String, schema: StructType)`, but I think that this introduces a small gap in functionality w.r.t the old code since I don't think `spark-redshift` exposes a great API to retrieve the schema of a Redshift table independent of unloading / reading that table, so it might be a little tricky for the user to actually obtain the `schema` object. We could try to address this by opening up an API for calling the schema-retrieval code over JDBC, but I don't think that is quite a sufficient solution since the old schemaString approach could be used in cases where the Spark cluster could not connect to Redshift / didn't require a JDBC connection. For this reason it might be nice to expose the `SchemaParser` object so that a user can call that as a shortcut helper method for constructing the Spark `Schema` object. However, the current `SchemaParser` is slightly painful to use from Java. Therefore, I think it might make sense to expose a public SchemaParser class in Java with static method and keep the parser implementation in a private Scala helper class that it calls.

I don't want to overengineer things now, though, so I'd be fine with rebasing and merging these changes as-is and dealing with feature-gaps / requests in followups or in a followup to the 2.x release.

JoshRosen:
Yeah, so TL;DR: rebase this and I'll merge it.

liancheng:
@JoshRosen Thanks! Having a Java version of `SchemaParser` sounds good to me. Finished rebasing.

JoshRosen:
LGTM so I'm going to merge this to master. Thanks @liancheng!

PR #253, opened 2016-08-05
Split COPY stmts and fix invalid s3 hostname exceptions with old aws-…

jameshou:
…java-sdk

We've seen a lot of messages lately regarding the "Invalid S3 URI: hostname does not appear to be a valid S3 endpoint" exception and so thought we should contribute our two cents and the code changes that worked for us.  We've tried many approaches listed in that thread including using spark.executor.extraClassPath and spark.driver.extraClassPath environment variables to prepend to the classpath, including it in the assembled jar or as a shaded jar,  Unfortunately many of these approaches failed mainly because we have on the machines themselves the aws-java-sdk jar and that usually takes precedence.  We ended up going with what Josh mentioned earlier about changing the s3 url in the spark-redshift code to add the endpoint to the host (*.s3.amazonaws.com).

Furthermore, we also ran into an issue where writing a large amount of data would fail (> 300GB of compressed data).  We found that we could upload the data to s3 but when we execute the COPY statement for a single manifest, the operation would timeout.  As such, we break out the manifest and run the COPY statement on each of the entry urls in the manifest.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -23,7 +23,7 @@ import java.net.URI
 import scala.collection.JavaConverters._
 
 import com.amazonaws.auth.AWSCredentials
-import com.amazonaws.services.s3.{AmazonS3Client, AmazonS3URI}
+import com.amazonaws.services.s3.AmazonS3Client
 import com.eclipsesource.json.Json
 import org.apache.spark.rdd.RDD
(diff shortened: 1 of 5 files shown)


JoshRosen:
Thanks for the PR. It looks like this PR incorporates two logically separate changes, the `COPY` change and the fix for the S3 hostnames. Would you mind separating these changes into two separate PRs so they can be reviewed and merged separately?

jameshou:
Thanks for getting back to us so quickly.  Sure will do.  I'll close this for now and reopen two separate PRs.

PR #254, opened 2016-08-05
Handle invalid S3 hostname exceptions with older aws-java-sdk versions

jameshou:
We've seen a lot of messages lately regarding the "Invalid S3 URI: hostname does not appear to be a valid S3 endpoint" exception and so thought we should contribute our two cents and the code changes that worked for us. We've tried many approaches listed in that thread including using `spark.executor.extraClassPath` and `spark.driver.extraClassPath` environment variables to prepend to the classpath, including it in the assembled jar or as a shaded jar, Unfortunately many of these approaches failed mainly because we have on the machines themselves the older aws-java-sdk jar and that usually takes precedence. We ended up going with what Josh mentioned earlier about changing the S3 url in the spark-redshift code to add the endpoint to the host (`*.s3.amazonaws.com`).

This logic will try to instantiate a new AmazonS3URI and if it fails, it'll try to add the default S3 Amazon domain to the host.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Utils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Utils.scala
@@ -80,7 +80,7 @@ private[redshift] object Utils {
    */
   def addEndpointToUrl(url: String, domain: String = "s3.amazonaws.com"): String = {
     val uri = new URI(url)
-    val hostWithEndpoint = uri.getHost() + "." + domain
+    val hostWithEndpoint = uri.getHost + "." + domain
     new URI(uri.getScheme,
       uri.getUserInfo,
       hostWithEndpoint,
--- a/src/main/scala/com/databricks/spark/redshift/Utils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Utils.scala
@@ -60,7 +60,8 @@ private[redshift] object Utils {
   }
 
   /**
-   * Factory method to create new S3URI in order to handle various library incompatabilities with older AWS Java Libraries
+   * Factory method to create new S3URI in order to handle various library incompatabilities with
+   * older AWS Java Libraries
    */
   def createS3URI(url: String): AmazonS3URI = {
     try {
@@ -68,7 +69,8 @@ private[redshift] object Utils {
       new AmazonS3URI(url)
     } catch {
       case e: java.lang.IllegalArgumentException => {
-        if (e.getMessage().startsWith("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint")) {
+        if (e.getMessage().
+          startsWith("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint")) {
           new AmazonS3URI(addEndpointToUrl(url))
         } else {
           throw e
--- a/src/main/scala/com/databricks/spark/redshift/Utils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Utils.scala
@@ -60,6 +60,40 @@ private[redshift] object Utils {
   }
 
   /**
+   * Factory method to create new S3URI in order to handle various library incompatabilities with older AWS Java Libraries
+   */
+  def createS3URI(url: String): AmazonS3URI = {
+    try {
+      // try to instantiate AmazonS3URI with url
+      new AmazonS3URI(url)
+    } catch {
+      case e: java.lang.IllegalArgumentException => {
+        if (e.getMessage().startsWith("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint")) {
+          new AmazonS3URI(addEndpointToUrl(url))
+        } else {
+          throw e
+        }
+      }
+    }
+  }
+
+  /**
+   * Since older AWS Java Libraries do not handle S3 urls that have just the bucket name
+   * as the host, add the endpoint to the host
+   */
+  def addEndpointToUrl(url: String, domain: String = "s3.amazonaws.com"): String = {
+    val uri = new URI(url)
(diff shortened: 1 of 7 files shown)


src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -60,6 +60,42 @@ private[redshift] object Utils {
...
+   * older AWS Java Libraries
+   */
+  def createS3URI(url: String): AmazonS3URI = {
+    try {
+      // try to instantiate AmazonS3URI with url
+      new AmazonS3URI(url)
+    } catch {
+      case e: java.lang.IllegalArgumentException => {
+        if (e.getMessage().
+          startsWith("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint")) {
```
JoshRosen: Is there a way to programatically get the AWS SDK version so that you don't have to use exceptions for control flow?

JoshRosen:
I'm going to test this out later by running all of the integration tests with an old AWS SDK. It might be cool to reconfigure the build so we can add an old SDK version to our test matrix. If you'd like to take a shot at this and don't mind SBT magic, take a look at how we handle the Spark and Avro versions in the build and test configurations (the key bit of trickiness is how we build against one version and test against another so that we catch binary compatibility problems)

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -60,6 +60,42 @@ private[redshift] object Utils {
...
+    }
+  }
+
+  /**
+   * Since older AWS Java Libraries do not handle S3 urls that have just the bucket name
+   * as the host, add the endpoint to the host
+   */
+  def addEndpointToUrl(url: String, domain: String = "s3.amazonaws.com"): String = {
+    val uri = new URI(url)
+    val hostWithEndpoint = uri.getHost() + "." + domain
```
JoshRosen: You can omit the `()` parents in these `get*()` calls; they're not necessary and cause style suggestions in IntelliJ.

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -60,6 +60,42 @@ private[redshift] object Utils {
   }
 
   /**
+   * Factory method to create new S3URI in order to handle various library incompatabilities with
```
JoshRosen: incompatibilities is misspelled here.

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -60,6 +60,42 @@ private[redshift] object Utils {
...
+   * older AWS Java Libraries
+   */
+  def createS3URI(url: String): AmazonS3URI = {
+    try {
+      // try to instantiate AmazonS3URI with url
+      new AmazonS3URI(url)
+    } catch {
+      case e: java.lang.IllegalArgumentException => {
```
JoshRosen: To avoid having to re-throw exceptions which don't have the correct message, you can combine the `if` with the `case` statement, e.g. `case e: IllegalArgumentException if e.getMessage.startsWith(...) =>`.

JoshRosen: AFAIK you also don't need to fully-qualify the exception.

src/main/scala/com/databricks/spark/redshift/Utils.scala
JoshRosen: It turns out that you can do this with https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/util/VersionInfoUtils.html#getVersion--, but it looks like the implementation of that method relies on being able to load a properties file from the SDK JAR and that might break under certain repackaging scenarios, so while a bit ugly I think this is fine for now. Just go ahead and clean things up so we don't have to explicitly re-throw.

JoshRosen:
I fear that this will break if we don't test it, so in https://github.com/databricks/spark-redshift/commit/39776a673e552e249579c5ba31a939d6a0a6b046 I've added end-to-end integration tests that run against the 1.7.4 version of the SDK. Could you go ahead and cherry-pick that change, address my comments, and push, then I'll retest and merge?

src/main/scala/com/databricks/spark/redshift/Utils.scala
```diff
@@ -60,6 +60,37 @@ private[redshift] object Utils {
...
   /**
+   * Factory method to create new S3URI in order to handle various library incompatibilities with
+   * older AWS Java Libraries
+   */
+  def createS3URI(url: String): AmazonS3URI = {
+    try {
+      // try to instantiate AmazonS3URI with url
+      new AmazonS3URI(url)
+    } catch {
+      case e: IllegalArgumentException if e.getMessage.startsWith("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint") => {
```
JoshRosen: Scalastyle is complaining that this line is too long. Try wrapping the string to the next line.

JoshRosen:
LGTM, so I'm going to merge this to master and will try to backport it to `branch-1.x`. Thanks @jameshou!

jameshou:
Yay!  Thanks for merging this in so quickly.  Can't wait to start using the official branch in our builds rather than including this in our patched up jar.  Let me know if there are any issues that come up and I will try to address them.

JoshRosen:
I'll try to package either a `2.0.1` or `2.1.0` release soon, but I'm currently blocking on resolving https://github.com/databricks/spark-avro/issues/147 so that I can bump the `spark-avro` version in that next release.

Issue #200, opened 2016-04-08
IAM Role credentials can expire between unload / load into Redshift


cmkimerer:
Note: This is likely superseded by #199, but I'll file in case this fix needs to be carried over, or #199 doesn't land in the next release.

When using `RedshiftWriter.saveToRedshift()` the credentials are fetched early in the function, if the `unloadData` call takes a significant amount of time (https://github.com/databricks/spark-redshift/blob/cba5eee1ab79ae8f0fa9e668373a54d2b5babf6b/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala#L374) it's possible that the credentials fetched have expired by the time `doRedshiftLoad` has run.

Ideally, the result returned from `loadFromUri` would return an instance of the InstanceProfileCredentialsProvider and getCredentials() would be called each time it was needed.  This would handle the fact that the credentials can expire and would re-fetch when necessary.  Obviously this change will break all other usages of fetching credentials right now, so perhaps an internal "credential wrapper" should be created to either return a wrapped BasicAWSCredentials when getCredentials() is called, or call through to the InstanceProfileCredentialsProvider.getCredentials() if that case is hit.

JoshRosen:
Just merged a patch addressing #199, so I think that hopefully your use-case should be addressed in the next release.

I do agree that we should be smarter about how we handle instance profile credentials, though. The solution that we have today is quasi-functional but places a lot of burden on the user for managing credential refresh, etc; it would be great to have the library handle more of this itself.

JoshRosen:
I've submitted #284 to fix this.

PR #277, opened 2016-09-29
Update AWSCredentialUtils to be uniform between URI schemes. 

a-morales:
Noticed that the way credentials were being obtained for `s3`/`s3n` were different than `s3a`. There shouldn't be different methods of getting credentials for the URI schemes as they should still be able to authenticate through the same means. 

This was causing me headaches as I tried to configure a spark job on EMR, but kept getting credential errors. After digging in the code and seeing the difference in how credentials are obtained, I tried using the `s3a` URI scheme for the temp path, which finally worked. Additionally the fallback to using `InstanceProfileCredentialsProvider` in the `s3a` method allowed me to not explicitly mess with credentials and just have the library "just work".


a-morales:
In terms of writing an integration test where using the `DefaultAWSCrendentialProviderChain` as an authentication method, the `IntegrationSuiteBase` will have to be updated to allow different methods of providing credentials as it always provides the credentials through the [hadoop config](https://github.com/databricks/spark-redshift/blob/master/src/it/scala/com/databricks/spark/redshift/IntegrationSuiteBase.scala#L89). 

This means that there is a probability that some of the integration tests for the other authentication methods are false positives, since they fall back to credentials in the hadoop configuration.

src/main/scala/com/databricks/spark/redshift/AWSCredentialsUtils.scala
```diff
@@ -71,8 +69,8 @@ private[redshift] object AWSCredentialsUtils {
           }
         }.orElse {
           // Next, try to read from configuration
-          val accessKey = hadoopConfiguration.get("fs.s3a.access.key", null)
-          val secretKey = hadoopConfiguration.get("fs.s3a.secret.key", null)
+          val accessKey = hadoopConfiguration.get(s"fs.$uriScheme.access.key", null)
```
JoshRosen: As far as I know `fs.s3n.access.key` won't work; S3A uses a different scheme for encoding credentials than the other file systems.

a-morales: This is addressed in commit 6f04487

JoshRosen:
I just tested this out and it worked beautifully. I have a couple of wording changes that I'd like to suggest for the docs, though, since I'd like to clarify the usage of `aws_iam_role`: if you're running on EMR or another environment with instance profile credentials then you may end up obtaining temporary credentials which may expire between the time that we get them and the time that we pass them to Redshift (since we might obtain the credentials, issue a very slow write to S3, then re-use now-expired credentials to issue load requests to Redshift). However, I think that we can mitigate this risk by fixing #200 and having the credentials be re-fetched on every usage.

I'm going to run the integration tests with this and will merge it provided that they pass, then I'll work on fixing #200 and will also submit a separate patch to update the docs to clarify the various use-cases for each authentication method. Thanks for your patch!

PR #282, opened 2016-10-17
Updated README.md with sample code to read a Redshift table in SparkR

ganeshchand:
Added sample code for reading a Redshift table with SparkR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,15 +186,24 @@ Reading data using R:
 **Spark 2.0**
 
 ```R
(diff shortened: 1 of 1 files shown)


JoshRosen:
This looks pretty good, but the style was slightly inconsistent with the existing examples so I went ahead and updated them. I'd also like to remove the separate Spark 1.6 and 2.x examples from the README since the current master README only applies to Spark 2.x (since Spark-redshift 2.x only supports Spark 2.x). Therefore I'd like to commit slightly different PRs against master and branch-1.x.

I'm going to make a quick update / edit to this now, then will commit this for master and will port a modified version to branch-1.x.

PR #284, opened 2016-10-18
Pass around AWSCredentialProviders instead of AWSCredentials

### LINKED ISSUE #200: IAM Role credentials can expire between unload / load into Redshift

JoshRosen:
This patch refactors the library internals to pass around `AWSCredentialProvider` instances instead of `AWSCredentials`, helping to avoid issues where temporary credentials are obtained at the start of a read or write operation and then expire when they are re-used later in the operation.

This fixes #200.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/AWSCredentialsUtils.scala
+++ b/src/main/scala/com/databricks/spark/redshift/AWSCredentialsUtils.scala
@@ -18,7 +18,7 @@ package com.databricks.spark.redshift
 
 import java.net.URI
 
-import com.amazonaws.auth.{BasicAWSCredentials, AWSCredentials, AWSSessionCredentials, DefaultAWSCredentialsProviderChain}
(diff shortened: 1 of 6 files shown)


JoshRosen:

PR #285, opened 2016-10-19
Add documentation and warnings related to using different regions for Redshift and S3

### LINKED ISSUE #87: Add configurations to allow tempdir and Redshift cluster to be in different AWS regions
By default, S3 <-> Redshift copies will not work if the S3 bucket and Redshift cluster are in different AWS regions. If you try to use a bucket in a different region, then you get a confusing error message; see https://forums.databricks.com/questions/1963/why-spark-redshift-can-not-write-s3-bucket.html for one example.

Note that it is technically possible to use a bucket in a different region if you pass an extra `region` parameter to the COPY command; see https://sqlhaven.wordpress.com/2014/09/07/common-errors-of-redshift-copy-command-and-how-to-solve-them-part-1/ for one example of this. 

~~~Although we _could_ support this, I don't think that it necessarily is a high priority / makes sense: there's no good reason that I can think of for a `spark-redshift` user to write short-lived, temporary data into a bucket in a different region and pay cross-region data transfer costs.~~~

~~As a result, I think that we should document this limitation and possibly add some configuration validation to print a better error message when the S3 bucket is in the wrong region.~~~

We should add a configuration option so that users can explicitly specify the `tempdir` region to enable cross-region copies.

JoshRosen:
Redshift throws extremely confusing errors when the Redshift cluster and S3 bucket are in different AWS regions. This patch expands the documentation to discuss how to fix these errors and adds logic to attempt to automatically detect and warn users when this case occurs.

Fixes #87.


JoshRosen:

PR #287, opened 2016-10-20
Use CSV data source to read unloaded data from Redshift

JoshRosen:
This patch refactors this library's read path to use Spark 2.0's built-in CSV data source to read unloaded Redshift output from S3. This approach has a few advantages over our existing custom-InputFormat-based code:

- It will benefit from performance improvements in `FileScanRDD` and `HadoopFsRelation`, including automatic coalescing.
- The Univocity CSV parser is likely to be faster than our old custom code.
- We don't have to create a separate RDD per partition and union them together, making the RDD DAG smaller.

In order to do this, I had to change the `UNLOAD` command to add the `ADDQUOTES` option. Previously, the unloaded data used `|` as a delimiter and used `\|` to escape that delimiter, but it appears that Spark's CSV data source isn't able to handle non-quoted escaping of delimiters (@falaki, could you confirm?). I believe that this should be safe, but it would be great to figure out why this library didn't take this approach in the first place. @jaley, do you know of any gotchas in doing this?

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Conversions.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Conversions.scala
@@ -78,7 +78,7 @@ private[redshift] object Conversions {
    *
    * Note that instances of this function are NOT thread-safe.
    */
-  def createRowConverter(schema: StructType): (Array[String]) => Row = {
+  def createRowConverter(schema: StructType): Row => Row = {
     val dateFormat = createRedshiftDateFormat()
     val decimalFormat = createRedshiftDecimalFormat()
     val conversionFunctions: Array[String => Any] = schema.fields.map { field =>
@@ -110,10 +110,10 @@ private[redshift] object Conversions {
     }
     // As a performance optimization, re-use the same mutable Seq:
     val converted: mutable.IndexedSeq[Any] = mutable.IndexedSeq.fill(schema.length)(null)
-    (fields: Array[String]) => {
+    (inputRow: Row) => {
       var i = 0
       while (i < schema.length) {
-        val data = fields(i)
+        val data = inputRow.getString(i)
         converted(i) = if (data == null || data.isEmpty) null else conversionFunctions(i)(data)
         i += 1
       }
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftRelation.scala
@@ -17,7 +17,6 @@
 package com.databricks.spark.redshift
 
 import java.io.InputStreamReader
-import java.lang
 import java.net.URI
 
 import scala.collection.JavaConverters._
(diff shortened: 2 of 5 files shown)


liancheng:
Put the newline issue aside, if we get this merged, we should be able to remove `RedshiftInputFormat`, right?

JoshRosen:
Yes, in principle we could remove `RedshiftInputFormat`. I think that it's _technically_ a public API in the sense that it's mentioned offhand in the README, but it's really hard to use and I doubt that users of that `InputFormat` are also using the more full-fledged functionality of this library as part of the same project. Therefore I'd be fine with removing it, especially if this version will become 3.0.0. I might take care of that in a separate patch, though.

JoshRosen:
Regarding newlines, I think that our use of the `ESCAPE` unload option should take care of this. From the [Redshift UNLOAD docs](https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html):

> ESCAPE
>
> For CHAR and VARCHAR columns in delimited unload files, an escape character (\) is placed before every occurrence of the following characters:

> - Linefeed: \n
> - Carriage return: \r
> - The delimiter character specified for the unloaded data.
> - The escape character: \
> - A quote character: " or ' (if both ESCAPE and ADDQUOTES are specified in the UNLOAD command).

Univocity's default behavior is to process escapes and expand them. It's probably a good idea to add some more end-to-end tests with this, though: I can add some tests to `RedshiftReadSuite` which write strings containing all of these special characters to Redshift and try to read them back as part of the same unload.

liancheng:
Just checked with @mengxr, since Redshift already supports Avro now, users should be able to workaround the newline issue (if it still exists now) by switching to Avro. Since we are about to bumping the version to 3.0.0, I think it's fine.

mengxr:
@JoshRosen When UNLOAD, redshift literally put a `\` before a newline character if it exists in the record, not replacing newline by `\n`. This leads to multi-line records, which were not supported by CSV source, at least at the time I wrote this input format. If I remember correctly, `ADDQUOTES` still leaves newline chars as is.

There is a test for this at https://github.com/databricks/spark-redshift/blob/master/src/test/scala/com/databricks/spark/redshift/RedshiftInputFormatSuite.scala#L69. You can check whether CSV source can load that dataset correctly.

If CSV source can handle those multi-line records correctly, we should remove the input format.

JoshRosen:
Let me check tomorrow.  Even if we can't use the CSV reader I think we'll still be able to refactor to use HadoopFSRelation with a custom FileFormat

JoshRosen:
@mengxr, I just tested this and you're right: Redshift is just prepending `\` rather than replacing the original characters with escapes. However, this appeared to work by accident for small input files because the Univocity parser supports multi-line records by default. It looks like Univocity has a `recordEndsOnNewline` option which can be set to enforce that requirement, which would help to guard against cases where multi-line records appear to work but will break as soon as files get large enough to be split.

@marmbrus, @falaki, @mengxr, do you think that we should enable `recordEndsOnNewline` to prevent users from writing code which works by accident on small files but fails in a large data environment? This risks breaking users' local programs which are just using Spark as a convenient programming API for small data, but could help to avoid user confusion. Probably fine to leave it as-is but we might want to add a warning to the open-source documentation.

JoshRosen:
I'd still like to use a FileFormat-based DataSource since that will handle auto-coalescing, so I'm going to close this PR and will open a new one which replaces the CSV usage with a custom format (which can hopefully be a thin harness wrapping the InputFormat).

marmbrus:
I would probably put a warning, rather than change the behavior.

mengxr:
Even Univocity supports multi-line records, we still need to keep the special logic to seek for the start position in a file split to avoid partial records near the split boundary. See https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/RedshiftInputFormat.scala#L168.

PR #289, opened 2016-10-25
Use FileFormat-based data source instead of HadoopRDD for reads

JoshRosen:
This patch refactors this library's read path to use a Spark 2.0's `FileFormat`-based data source to read unloaded Redshift output from S3. This approach has a few advantages over using our existing `HadoopRDD`-based approach:

- It will benefit from performance improvements in `FileScanRDD` and `HadoopFsRelation`, including automatic coalescing.
- We don't have to create a separate RDD per partition and union them together, making the RDD DAG smaller.

The bulk of the diff are helper classes copied from Spark and `spark-avro` and inlined here for API compatibility / stability purposes. Some of the new classes implemented here are likely to become incompatible with new releases of Spark, but note that `spark-avro` itself relies on similar unstable / experimental APIs and thus this library is already vulnerable to changes to those APIs (in other words, this change is not making our compatibility story significantly worse).

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RecordReaderIterator.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RecordReaderIterator.scala
@@ -20,13 +20,9 @@ package com.databricks.spark.redshift
 import java.io.Closeable
 
 import org.apache.hadoop.mapreduce.RecordReader
-import org.apache.spark.sql.catalyst.InternalRow
 
 /**
  * An adaptor from a Hadoop [[RecordReader]] to an [[Iterator]] over the values returned.
- *
- * Note that this returns [[Object]]s instead of [[InternalRow]] because we rely on erasure to pass
- * column batches by pretending they are rows.
  *
  * This is copied from Apache Spark and is inlined here to avoid depending on Spark internals
  * in this external library.
--- a/src/main/scala/com/databricks/spark/redshift/RecordReaderIterator.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RecordReaderIterator.scala
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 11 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftFileFormat.scala
```diff
@@ -0,0 +1,105 @@
...
+      options: Map[String, String],
+      dataSchema: StructType): OutputWriterFactory = {
+    throw new UnsupportedOperationException(s"prepareWrite is not supported for $this")
+  }
+
+  override def isSplitable(
+      sparkSession: SparkSession,
+      options: Map[String, String],
+      path: Path): Boolean = {
+    // Redshift unload files are not splittable because records containing newline characters may
```
JoshRosen: Per discussion with @mengxr, it sounds like the existing `InputFormat` already does support splitting, so this comment is incorrect. I'll go ahead and update this.

src/main/scala/com/databricks/spark/redshift/RecordReaderIterator.scala
```diff
@@ -0,0 +1,74 @@
...
+import java.io.Closeable
+
+import org.apache.hadoop.mapreduce.RecordReader
+import org.apache.spark.sql.catalyst.InternalRow
+
+/**
+ * An adaptor from a Hadoop [[RecordReader]] to an [[Iterator]] over the values returned.
+ *
+ * Note that this returns [[Object]]s instead of [[InternalRow]] because we rely on erasure to pass
+ * column batches by pretending they are rows.
```
liancheng: This paragraph comes from Spark source code and doesn't really apply here. It confused me for a while.

JoshRosen: Yeah, this is confusing to me as well, especially since the return type is generic. I'll just remove this comment.

liancheng:
LGTM!

JoshRosen:
Great, merging to master!

PR #291, opened 2016-10-26
Make Redshift to S3 authentication mechanisms mutually exclusive

JoshRosen:
This patch makes a breaking change to how Redshift to S3 communication is authenticated. Previously, the implicit default behavior was to forward Spark's S3 credentials to Redshift and this default would be used unless `aws_iam_role` or the `temporary_aws_*` options were set. This behavior is slightly dodgy because it meant that a typo in the IAM settings (i.e. using the parameter `redshift_iam_role` instead of the correct `aws_iam_role`) would cause a default authentication mechanism to be used instead.

To fix that gap, this patch changes this library so that Spark's S3 credentials will only be forwarded to Redshift if `forward_spark_s3_credentials` is set to `true`. This option is mutually-exclusive with the `aws_iam_role` and `temporary_aws_*` options and is set to `false` by default. The net effect of this change is that users who were already using ``aws_iam_role` or the `temporary_aws_*` options will be unaffected, while users relying on the old default behavior will need to set `forward_spark_s3_credentials` to `true` in order to continue using that authentication scheme.

I have updated the README with a new section explaining the different connections involved in using this library and the different authentication mechanisms available for them. I also added a new section describing how to enable encryption of data in transit and at rest.

Because of the backwards-incompatible nature of this change, I'm bumping the version number to `3.0.0-preview1-SNAPSHOT`.


yhuai:
lgtm1

README.md
```diff
@@ -211,42 +212,163 @@ val records = sc.newAPIHadoopFile(
...
+        sc._jsc.hadoopConfiguration().set("fs.s3n.awsAccessKeyId", "YOUR_KEY_ID")
+        sc._jsc.hadoopConfiguration().set("fs.s3n.awsSecretAccessKey", "YOUR_SECRET_ACCESS_KEY")
+        ```
+
+    3. **Encode keys in `tempdir` URI**:
+     For example, the URI `s3n://ACCESSKEY:SECRETKEY@bucket/path/to/temp/dir` encodes the key pair
+      (`ACCESSKEY`, `SECRETKEY`).
+
+      Due to [Hadoop limitations](https://issues.apache.org/jira/browse/HADOOP-3733), this
+      approach will not work for secret keys which contain forward slash (`/`) characters.
```
thomaso-mirodin: I assume most users are familiar with the need to urlencode s3 urls as the workaround to this type of issue; however since that doesn't work maybe mention that urlencode won't fix this issue (for people like me who didn't open the link on the first read through :D)

README.md
```diff
@@ -211,42 +212,163 @@ val records = sc.newAPIHadoopFile(
...
+        sc.hadoopConfiguration.set("fs.s3a.secret.key", "YOUR_SECRET_ACCESS_KEY")
+        ```
+
+        Python users will have to use a slightly different method to modify the `hadoopConfiguration`,
+        since this field is not exposed in all versions of PySpark. Although the following command
+        relies on some Spark internals, it should work with all PySpark versions and is unlikely to
+        break or change in the future:
+
+        ```python
+        sc._jsc.hadoopConfiguration().set("fs.s3n.awsAccessKeyId", "YOUR_KEY_ID")
```
thomaso-mirodin: Is there a non-internal method that may not be as pretty or is this the best we can do?

i.e. for pyspark > N use this supported method, for pyspark <= N use this workaround

README.md
```diff
@@ -211,42 +212,163 @@ val records = sc.newAPIHadoopFile(
...
+        guide to associate that IAM role with your Redshift cluster.
+        4. Set this library's `aws_iam_role` option to the role's ARN.
+    2. **Forward Spark's S3 credentials to Redshift**: if the `forward_spark_s3_credentials` option is
+        set to `true` then this library will automatically discover the credentials that Spark is
+        using to connect to S3 and will forward those credentials to Redshift over JDBC. If Spark
+        is authenticating to S3 using an IAM instance role then a set of temporary STS credentials
+        will be passed to Redshift; otherwise, AWS keys will be passed. These credentials are
+        sent as part of the JDBC query, so therefore it is **strongly recommended** to enable SSL
+        encryption of the JDBC connection when using this authentication method.
+    3. **Use Security Token Service (STS) credentials**: You may configure the
```
thomaso-mirodin: These are also passed from the driver to redshift right? If so, shouldn't we also be encouraging TLS?

README.md
```diff
@@ -211,42 +212,163 @@ val records = sc.newAPIHadoopFile(
...
-5. **IAM instance profiles:** If you are running on EC2 and authenticate to S3 using IAM and [instance profiles](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html), then you must must configure the `temporary_aws_access_key_id`, `temporary_aws_secret_access_key`, and `temporary_aws_session_token` configuration properties to point to temporary keys created via the AWS [Security Token Service](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html). These temporary keys will then be passed to Redshift via `LOAD` and `UNLOAD` commands.
+This library reads and writes data to S3 when transferring data to/from Redshift. As a result, it
+requires AWS credentials with read and write access to a S3 bucket (specified using the `tempdir`
+configuration parameter).
+
+> **:warning: Note**: This library does not clean up the temporary files that it creates in S3.
```
thomaso-mirodin: Do these temporary files also need to be encrypted & authenticated and thus touched on in the encryption section?

README.md
thomaso-mirodin: If they're passed through some other mechanism than JDBC, maybe we can document that as well?

README.md
JoshRosen: Alas, there's still not a better method than this :(

README.md
JoshRosen: These are the temporary files being referred to in the _Encrypting `UNLOAD` data stored in S3 (data stored when reading from Redshift)_ and _Encrypting `COPY` data stored in S3 (data stored when writing to Redshift)_ headings under the _Encryption_ setting.

README.md
JoshRosen: Done.

thomaso-mirodin:
Lgtm

JoshRosen:
I'm going to merge this and will begin packaging a preview release. Thanks for the reviews!

steveloughran:
One thing to note here is that s3a in Hadoop 2.8 supports STS via the property `fs.s3a.session.token`; to be ready for that Hadoop version you should really be syncing that too. 

See: [hadoop-aws docs](https://github.com/apache/hadoop/blob/branch-2/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md)

JoshRosen:
Good catch @steveloughran; I've filed #296 to make sure that I don't forget to fix this before the next release.

PR #301, opened 2016-11-16
Use PreparedStatement.getMetaData() to retrieve Redshift query schemas

### LINKED ISSUE #300: Error from Redshift when creating DataFrame from result of 'approximate count(distinct)' query
Attempting to create a dataframe from the result of a Redshift query containing an `approximate count(distinct ...)` column, such as

```
    read.option("query", s"select approximate count(distinct testbool) from $test_table").load()
```

would result in an error like

```
java.sql.SQLException: [Amazon](500310) Invalid operation: approximate count(distinct) must be applied on user tables;
```

while trying to resolve the schema of the query result.

This is caused by the way that this library discover's a query's result schema by issuing queries like

```
SELECT * FROM (select approximate count(distinct testbool) from read_suite_test_table_5263604562389761851) WHERE 1=0
```

which Redshift doesn't understand.

This bug can be fixed via a small change to our table resolution code.

JoshRosen:
Previously, this library issued a query against Redshift and inspected the ResultSet's metadata to determine the column names and types. Unfortunately, this approach does not work in a few corner-cases, including retrieving the schema of queries which use the `approximate count(distinct ...)` construct.

This patch fixes this issue by modifying our code to call `getMetaData()` on the prepared statement rather than running the query.   

Fixes #300.

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -151,9 +151,9 @@ private[redshift] class JDBCWrapper {
    * @throws SQLException if the table contains an unsupported type.
    */
   def resolveTable(conn: Connection, table: String): StructType = {
-    val rs = executeQueryInterruptibly(conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0"))
+    val ps = conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0")
     try {
-      val rsmd = rs.getMetaData
+      val rsmd = executeInterruptibly(ps, _.getMetaData)
       val ncols = rsmd.getColumnCount
       val fields = new Array[StructField](ncols)
       var i = 0
@@ -170,7 +170,7 @@ private[redshift] class JDBCWrapper {
       }
       new StructType(fields)
     } finally {
-      rs.close()
+      ps.close()
     }
   }
 
@@ -267,7 +267,8 @@ private[redshift] class JDBCWrapper {
     // Somewhat hacky, but there isn't a good way to identify whether a table exists for all
     // SQL database systems, considering "table" could also include the database name.
     Try {
-      executeQueryInterruptibly(conn.prepareStatement(s"SELECT 1 FROM $table LIMIT 1")).next()
+      val stmt = conn.prepareStatement(s"SELECT 1 FROM $table LIMIT 1")
+      executeInterruptibly(stmt, _.getMetaData).getColumnCount
     }.isSuccess
   }
 
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -151,6 +151,10 @@ private[redshift] class JDBCWrapper {
    * @throws SQLException if the table contains an unsupported type.
    */
   def resolveTable(conn: Connection, table: String): StructType = {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -151,9 +151,9 @@ private[redshift] class JDBCWrapper {
    * @throws SQLException if the table contains an unsupported type.
    */
   def resolveTable(conn: Connection, table: String): StructType = {
-    val rs = executeQueryInterruptibly(conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0"))
+    val ps = conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0")
```
JoshRosen: Note: it's still important to leave the `WHERE 1=0` clause in order to limit the work of the query in case the underlying JDBC driver implementation happens to implement this by actually executing the query (some abstract JDBC driver implementations do this; it looks like Redshift and Postgres don't but I'd like to leave this code here as a safety-net in case older drivers do something silly).

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -267,7 +267,8 @@ private[redshift] class JDBCWrapper {
     // Somewhat hacky, but there isn't a good way to identify whether a table exists for all
     // SQL database systems, considering "table" could also include the database name.
     Try {
-      executeQueryInterruptibly(conn.prepareStatement(s"SELECT 1 FROM $table LIMIT 1")).next()
+      val stmt = conn.prepareStatement(s"SELECT 1 FROM $table LIMIT 1")
```
JoshRosen: If `$table` happens to contain an expensive-to-compute scalar subquery then this might be an expensive operation, so I've updated this to attempt to avoid execution using the same `.getMetaData()` trick.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
```diff
@@ -151,9 +151,9 @@ private[redshift] class JDBCWrapper {
    * @throws SQLException if the table contains an unsupported type.
    */
   def resolveTable(conn: Connection, table: String): StructType = {
-    val rs = executeQueryInterruptibly(conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0"))
+    val ps = conn.prepareStatement(s"SELECT * FROM $table WHERE 1=0")
     try {
-      val rsmd = rs.getMetaData
+      val rsmd = executeInterruptibly(ps, _.getMetaData)
```
JoshRosen: As a defensive measure, we need to guard against interrupt-swallowing here.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
yhuai: Seems we should put this comment in the code?

JoshRosen:
By the way, I think the changes here should end up unblocking my concern in #204, so I'll update and merge a new version of that patch after this goes in.

yhuai:
lgtm

JoshRosen:
The test failure here is unrelated (something is flaky at the project infra level since I think our code coverage reporter might be OOM'ing or something), so I'm going to merge this and will update the RADME to explain how to download snapshots containing this fix.

src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
JoshRosen: Done.

Issue #310, opened 2016-12-09
Redshift Bigint may resolve to DoubleType or Long


ngerstle:
A query to redshift (example:
```val res =  queryFunc("select count(column.name) as notnull from features_permanent")``` )
may return either a BigInt (resolving as Long) or Double (resolving as Double).

This may derive from code copied in from spark in commit 2cd12e7, and seems to derive from https://github.com/apache/spark/commit/a1e092eae57172909ff2af06d8b461742595734c, which assigns type based on whether the result is signed. 

While this may be a proper result for generic jdbc connected database bigints (arguable), redshift does not have unsigned bigints, and redshift bigints should be resolved to longs.
Workaround are possible based on detecting type at runtime, but are ugly.

A pull request fixing https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala#L304 will follow shortly.

ngerstle:
It may be that `select count(columnname)` returns a `double` rather than a `bigint` when the column type is `double`. Still testing/checking.

ngerstle:
I can confirm that of 293ish columns, 178 of 178 columns of type `double` in redshift return a `double` when a query of `select count(doublecolumn)::BIGINT from table`.

PR #311, opened 2016-12-09
Update RedshiftJDBCWrapper.scala

ngerstle:
java.sql.Types.BIGINT => LongType in all cases for Redshift.
Redshift has no unsigned BigInt, and thus BigInt maps to Long without loss of resolution.
See issue #310

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
+++ b/src/main/scala/com/databricks/spark/redshift/RedshiftJDBCWrapper.scala
@@ -301,7 +301,7 @@ private[redshift] class JDBCWrapper {
     val answer = sqlType match {
       // scalastyle:off
       case java.sql.Types.ARRAY         => null
-      case java.sql.Types.BIGINT        => if (signed) { LongType } else { DecimalType(20,0) }
+      case java.sql.Types.BIGINT        => LongType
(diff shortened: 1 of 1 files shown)


JoshRosen:
Thanks for the fix! Mind adding a regression test to an existing suite? If
you can't, I can tackle it in a followup. Code from your original issue
would be a good foundation for the test case. Check out RedshiftReadSuite,
since that's where I'd add it.

ngerstle:
I'm not able to find any clear point at which a decimal would be converted to a double, meaning that issue #310 is unlikely to fixed by this, but the dead code should still be pruned.
I added a couple of rough regression tests- please flesh out as recommended..

JoshRosen:
Regarding the double issue, is there a possibility that our call to ask Redshift the schema is returning a Double but then Redshift is actually turning around and unexpectedly giving us a Decimal? Do you have any test case which reproduces the Double behavior, even if you don't know how to actually fix the bug?

ngerstle:
Unfortunately, I'm still trying to get a consistent abstract case- as mentioned in #310, I have a `select count(column), count(distinct(column) from table` that iterates over columns: sometimes it returns `(bigint, bigint)` and sometimes `(double, bigint)` (when debugging the schema of the dataframe returned when using the query option on `sqlContext.read....`

PR #324, opened 2017-01-25
Add support for search_path

### LINKED ISSUE #295: TableName assumes PUBLIC schema when no schema is set
https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/TableName.scala#L42

Shouldn't this follow the rules of Redshift and respect the search_path of the logged in user:

http://docs.aws.amazon.com/redshift/latest/dg/r_search_path.html

sboora:
Hi @JoshRosen,

Added support to specify the Redshift schema and do not default to `PUBLIC` for non-schema-qualified table names.

Fixes: https://github.com/databricks/spark-redshift/issues/295


sboora:
ping! Highly appreciate if you can take a look soon.

viesti:
Any progress on this? 

We are using a different schema per environment but would like to keep the sql scripts not repeating the schema (or have them parameterized). This PR would quite probably solve our problem, so would be neat to get this merge&released :)

sboora:
@JoshRosen can you please take a look.

sboora:
We created a work around for this by creating Redshift Users with same name as the schema name, since Redshift defaults `'$user'` for `search_path`

I think it would still be valuable to merge this.

Gauravshah:
👍

Gauravshah:
have similar use case, anything I do to help this pull request ?

eriklovdahl:
Any progress on this?

PR #343, opened 2017-04-28
Include Column List option

### LINKED ISSUE #340: COPY command relies on column order in DataFrame
The write operation currently generates a `COPY` command like the following:

```sql
COPY "PUBLIC"."some_table" FROM 's3://some-bucket/tmp/manifest.json' CREDENTIALS 'aws_access_key_id=__;aws_secret_access_key=__' FORMAT AS CSV NULL AS '@NULL@' manifest 
```

This relies on the DataFrame to have the columns in the same order as the table if it already exists. However, the `COPY` command supports specifying column lists or JSONPath expressions to map columns ([documentation](http://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-column-mapping.html)). It would be nice to at least support the column list, potentially as an option on the write operation like:

```scala
df.write
  .option("url", "jdbc:redshift://redshifthost:5439/database?user=username&password=pass")
  .option("dbtable", "my_table_copy")
  .option("tempdir", "s3n://path/for/temp/data")
  .option("include_column_list", "true")
  .mode("error")
  .save()
```

Looks like this should be fairly straightforward to add [here](https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/RedshiftWriter.scala#L89-L103).

kyrozetera:
Closes https://github.com/databricks/spark-redshift/issues/340

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,6 +571,16 @@ must also set a distribution key with the <tt>distkey</tt> option.
     <p>Since setting <tt>usestagingtable=false</tt> operation risks data loss / unavailability, we have chosen to deprecate it in favor of requiring users to manually drop the destination table themselves.</p>
     </td>
  </tr>
+  <tr>
+    <td><tt>include_column_list</tt></td>
+    <td>No</td>
+    <td>false</td>
+    <td>
+        If <tt>true</tt> then this library will automatically extract the columns from the schema
+        and add them to the COPY command according to the <a href="http://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-column-mapping.html">Column List docs</a>.
+        (e.g. `COPY "PUBLIC"."tablename" ("column1" [,"column2", ...])`).
+    </td>
+  </tr>
  <tr>
     <td><tt>description</tt></td>
     <td>No</td>
--- a/src/main/scala/com/databricks/spark/redshift/Parameters.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Parameters.scala
@@ -289,7 +289,7 @@ private[redshift] object Parameters {
 
     /**
      * If true then this library will extract the column list from the schema to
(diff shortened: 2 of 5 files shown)


frankfarrell:
This looks really useful. It would be really handy if we could have an option to ignore certain columns, eg columns that use IDENTITY, or columns with default values such as timestamps. That would solve https://github.com/databricks/spark-redshift/issues/381

toddwildey:
This would be very useful for our purposes and would make loading data into Redshift using this package far more robust for us.

kyrozetera:
@toddwildey Unfortunately this package is dead and they haven't accepted or merged any PRs on it in 2 years. We've decided to decouple our Spark processes from Redshift and handle the Redshift data access in a different layer and move away from this package since it's no longer maintained and got no traction from what I could tell on a potential fork.

eeshugerman:
Hi @kyrozetera I'm using [a (maintained) fork of this library](https://github.com/spark-redshift-community/spark-redshift) and this patch is just what I need. Would you be interested in contributing this patch to the fork? It applies without issue once the file paths are updated (see tweaked diff attached). If you'd rather not be bothered, I'd be happy to make the contribution on your behalf.

[columns-list-patch.txt](https://github.com/databricks/spark-redshift/files/3930020/columns-list-patch.txt)

kyrozetera:
@eeshugerman Sorry, I was out of town for the weekend so wasn't able to look at this until now. Looks like you've made the PR though so :+1:

sidrahsayyad:
Which version of the package can this change be found in ?

eeshugerman:
@sidrahsayyad https://github.com/spark-redshift-community/spark-redshift

sidrahsayyad:
I'm using pyspark --packages io.github.spark-redshift-community:spark-redshift_2.11:4.0.1 to test this change but consistently getting "Delimiter not found" error logged in stl_load_errors.

eeshugerman:
Did you set `include_column_list` to true? It's false by default.

PR #352, opened 2017-06-06
add 'create_table_if_not_exist' params

Joongho:
This is useful when merge operation.

1. create temp table A  (like schema.A)    -> run by preactions
2. load into table A ( dbtable = A )
3. update/insert     -> run by postactions

Above scenario, table A(=dbtable)  should not be created

### DIFF
--- a/src/main/scala/com/databricks/spark/redshift/Parameters.scala
+++ b/src/main/scala/com/databricks/spark/redshift/Parameters.scala
@@ -38,7 +38,8 @@ private[redshift] object Parameters {
     "diststyle" -> "EVEN",
     "usestagingtable" -> "true",
     "preactions" -> ";",
-    "postactions" -> ";"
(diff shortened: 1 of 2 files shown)


kyrozetera:
How is this different than setting a [SaveMode](https://spark.apache.org/docs/2.1.1/api/java/org/apache/spark/sql/SaveMode.html)?

Joongho:
It's upsert(=merge) operation.

schema.A is final target table,  A is temp table.

1. update schema.A set ... where schema.A.pk1 = A.pk1 and schema.A.pk2 = A.pk2 ... 
2. insert int schema.A 
           select * from A s 
                   left join schema.A t 
                    on t.pk1=s.pk1  and t.pk2=s.pk2
           where t.pk1 is null and t.pk2 is null ...

PR #367, opened 2017-10-11
Adding support for TIMESTAMPTZ as a Timestamp field to support the Am…

jontradesy:
Adding support for Amazon Redshift JDBC Driver 1.2.8. 

Redshift JDBC drivers of version 1.2.8 and later now treat the `TIMESTAMPTZ` (timestamp with a timezone) data type as a `TIMESTAMP` instead of a `VARCHAR` ([release notes here](https://s3.amazonaws.com/redshift-downloads/drivers/Amazon+Redshift+JDBC+Release+Notes.pdf)). The current timestamp parser cannot handle timestamps with timezone information (ex. `2017-09-22 14:03:11.503+03`). This fix will parse timestamps with a timezone and normalize them to UTC time. For example the `TIMESTAMPTZ` value `2017-09-22 14:03:11.503+03` will be normalized to `2017-09-22 11:03:11.503` (note the 3 hour shift).

## Test Updates
The tests now use the latest JDBC Redshift driver as defined by 
```scala
"com.amazon.redshift" % "redshift-jdbc42" % "1.2.8.1005"
```
Test cases have been added for `TIMESTAMPTZ` fields

## Versioning
The inherent backwards incompatibility of the Redshift JDBC Driver 1.2.8 suggests that this update should be a major release as defined by [semantic versioning](http://semver.org/).

The Migration Guide in the README is updated to include the changes necessary to use the new version.


ahharu:
please!!! this is breaking our AWS Glue code all of a sudden!!

